PR #2, opened 2025-03-14
add workflow

jyungtong:


.tool-versions
```diff
@@ -1 +1 @@
-nodejs 20.10.0
\ No newline at end of file
+nodejs 22.14.0
```
jyungtong, line 1: 22 is the next LTS

jyungtong:
will have another PR to do the centralised thing

package.json
```diff
@@ -1,10 +1,14 @@
...
+  "name": "@monorise",
+  "version": "0.0.0",
   "description": "",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
+    "build": "tsc",
+    "release": "npm run build && npx @changesets/cli publish",
+    "version": "npx @changesets/cli version",
+    "changeset": "npx @changesets/cli",
```
wengkhing, line 11: trying to understand the difference between a normal npm publish and changeset publish. For changeset, it additionally also manage releases and tag ya?

jyungtong: yes tags will be automatically managed

PR #54, opened 2025-04-10
update local mutual state in editEntity and some minor fix

yeuharn96:
- `editEntity`: update local mutual state to latest entity data
- `useEntities`: expose `lastKey` & `isFirstFetched` attribute
- `useMutuals`: expose `lastKey` attribute and added `listMore` function
- expose `TagRepository` type
- `listEntitiesByEntity`: added `'#'` at the end of `SK` value to prevent accidentally got unwanted entity (eg.: desire to get `company` entity but returned both `company` & `company-staff` entities)

### DIFF
--- a/packages/react/actions/core.action.ts
+++ b/packages/react/actions/core.action.ts
@@ -272,9 +272,10 @@ const initCoreActions = (
       produce((state) => {
         state.entity[entityType].dataMap.set(data.entityId, data);
 
+        // update mutual's entity data
         for (const key of Object.keys(state.mutual)) {
-          const [byEntity, entity, byId] = key.split('/');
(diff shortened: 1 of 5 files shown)


packages/react/actions/core.action.ts
```diff
@@ -271,6 +271,17 @@ const initCoreActions = (
     monoriseStore.setState(
       produce((state) => {
         state.entity[entityType].dataMap.set(data.entityId, data);
+
+        for (const key of Object.keys(state.mutual)) {
+          const [byEntity, entity, byId] = key.split('/');
```
jyungtong, line 276:
```suggestion
          const [_byEntity, entity, _byId] = key.split('/');
```
for clarity

PR #57, opened 2025-04-11
chore(core): add tests for Entity and EntityRepository

jyungtong:


jyungtong:
#59

wengkhing:
Thinking if we could also include a badge like this for test passed for production package: 
image

wengkhing (approved):

jyungtong:
> Thinking if we could also include a badge like this for test passed for production package: image

Good idea, i think can do when we improving the docs.

PR #62, opened 2025-04-11
fix mutual state key

yeuharn96:
## Issue

image
when we are using `useMutuals` to list data, we are listing from state key `job-application/activity/.../list`, however when we `createMutual`, the new mutual data is stored inside state key `job-application/.../activity`, meaning that `useMutuals` hook will not recognize the latest created mutual because it is stored inside a different state key

## Fix

change all mutual related state action to use same state key pattern

### DIFF
--- a/packages/react/actions/core.action.ts
+++ b/packages/react/actions/core.action.ts
@@ -12,7 +12,7 @@ import {
   constructMutual,
   flipMutual,
 } from '../lib/entity';
-import { convertToMap } from '../lib/utils';
+import { convertToMap, mutualStateKey, tagStateKey } from '../lib/utils';
 import type {
   CommonOptions,
   CoreService,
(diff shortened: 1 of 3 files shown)


yeuharn96:
refer to this PR https://github.com/monorist/monorise/pull/40

jyungtong:
> refer to this PR #40
> 
> i suggest we can create a helper function for all state keys instead of hard code it everytime

@yeuharn96 good idea, do you think it's possible we take this PR as a chance to refactor?

packages/react/actions/core.action.ts
```diff
@@ -314,7 +314,8 @@ const initCoreActions = (
     opts: CommonOptions = {},
     chainEntityQuery?: string,
   ) => {
-    const selfKey = opts.stateKey ?? `${byEntityType}/${id}/${entityType}`;
+    const selfKey =
+      opts.stateKey ?? mutualStateKey(byEntityType, id, entityType);
```
jyungtong, line 318: `chainEntityQuery` should pass in right?

packages/react/lib/utils.ts
```diff
@@ -10,3 +12,20 @@ export const convertToMap = <T extends Record<string, any>>(
...
   return map;
 };
+
+export const mutualStateKey = (
+  byEntity: Entity,
+  byEntityId: string | null,
+  entity: Entity,
+  chainEntityQuery?: string,
+) => {
+  return `${byEntity}/${byEntityId}/${entity}${chainEntityQuery ? `?${chainEntityQuery}` : ''}`;
```
jyungtong, line 22: do you think we can also handle `entityId` here?

yeuharn96: [c3a5085](https://github.com/monorist/monorise/pull/62/commits/c3a5085c90097aeb4b4b5df7e76f0d225d8efbb5)

jyungtong (approved):

PR #64, opened 2025-04-24
improve concept doc

wengkhing:
Add definition and examples for each concepts

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 - [Monorise](#monorise)
   - [🚀 Features](#🚀-features)
-  - [Core Concepts](#core-concepts)
+  - [💡 Core Concepts](#💡-core-concepts)
   - [📦 Installation](#📦-installation)
     - [Using npm / yarn / etc](#using-npm-yarn-etc)
     - [Or clone the repo](#or-clone-the-repo)
@@ -34,7 +34,7 @@ Simplify Single table design with DynamoDB
 - Relationships management between entities
 - No more writing complex DynamoDB queries
 
-## Core Concepts
+## 💡 Core Concepts
 
 - Entity
 - Mutual
--- a/README.md
+++ b/README.md
@@ -1,5 +1,30 @@
 # Monorise
 
+<!--toc:start-->
+
+- [Monorise](#monorise)
+  - [🚀 Features](#🚀-features)
+  - [Core Concepts](#core-concepts)
+  - [📦 Installation](#📦-installation)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -33,13 +36,15 @@ npm install
 ## 🛠️ Usage
 
 ### Basic Example
+
 ```js
-import { yourFunction } from 'your-package-name'
+import { yourFunction } from "your-package-name";
 
-yourFunction('example')
+yourFunction("example");
```
jyungtong, line 43: based on our biome.json, we are using single quote tho. any reason we are changing this.

docs/CONCEPT.MD
```diff
@@ -1,21 +1,101 @@
...
-- **Tag**: To get a list of organizations marked as VIP, use `/tag/organizations/vip`
+---
 
-The above pattern remains the same for all CRUD operations for `Entity` and `Mutual` endpoints.
\ No newline at end of file
+# Definitions ✍️
+
+This system revolves around three core building blocks: `Entity`, `Mutual`, and `Tag`.
+
+Want a look under the hood? Check out all the API routes [here](https://github.com/monorist/monorise/blob/4de00a930b5143b9365970fcfdb9a1456ae5eb98/packages/core/controllers/setupRoutes.ts) for the full picture.
```
jyungtong, line 23:
```suggestion
Want a look under the hood? Check out all the API routes [here](https://github.com/monorist/monorise/blob/main/packages/core/controllers/setupRoutes.ts) for the full picture.
```
so it always point to latest file

docs/CONCEPT.MD
```diff
@@ -1,21 +1,91 @@
...
+- Tags can be queried by **key, value, or both**
+- Structured tags (e.g., `priority#high`, `createdAt#2025-04-24`) can be **sorted** or used in range queries
+
+#### Example 📑
+
+Imagine an **organization** entity with the following tags:
+
+- `region#eu-west-1`: `activatedAt#2025-05-01`
+- `<empty>`: `activatedAt#2025-05-01`
+- `status#active`: `<empty>`
```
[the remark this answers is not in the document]
wengkhing: I'm thinking to explain this in a separate API doc which will explain each API function more in-depth. I think the concept doc is serve as an overview but not for API explanation.

README.md
wengkhing: it's auto corrected by my markdown formatter, didn't notice about this, will revert back

jyungtong (approved):

PR #67, opened 2025-04-24
Feat/add unique fields to entity config

wengkhing:
## What

Update `createEntityConfig` to accept `uniqueFields`. Currently, it does nothing, but this would enable update of core package to handle the following proposed behaviour.

### Configuration

Make sure that the fields stated in `uniqueFields` are also defined in `baseSchema`, similar to `searchableFields`.

```js
{
  // createEntityConfig
  baseSchema: z.object({
    username: z.string(),
  }).partial(),
  uniqueFields: ['username'], // optional
}
```

### Behaviour

- If `uniqueFields` is specified, entity creation and updates will include a check to ensure that all values of the listed fields are unique within the scope of that entity type.
- If value already exists, creation/updates would throw validation error, eg. `username 'johndoe' already exists`.
- Unique value can also serve as a entry point to retrieve entity, `getEntityByUniqueField(Entity.User, 'username', 'johndoe')`
- On entity deletion, any associated unique field values will be released, making them available for reuse.

### DIFF
--- a/.changeset/solid-ducks-like.md
+++ b/.changeset/solid-ducks-like.md
@@ -1,0 +1,5 @@
+---
+"@monorise/base": patch
+---
+
+add unique fields to createEntityConfig
--- a/packages/base/types/monorise.type.ts
+++ b/packages/base/types/monorise.type.ts
@@ -90,6 +90,7 @@ export interface MonoriseEntityConfig<
    */
   createSchema?: CO;
   searchableFields?: (keyof B)[];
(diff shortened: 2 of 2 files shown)


wengkhing:
> Do you plan to use dynamodb's transactwrite to handle this? Or how's your checking flow will looks like?

Yup, I think it should use transactwrite to handle this.

yeuharn96 (approved):

PR #69, opened 2025-04-25
pass file type when upload and add option to disable loading

jaydenpung:
- add option to disable isInterruptive so I can not show loading bar
- pass file type to presign end point so I can use it as content type when creating presign url

### DIFF
--- a/packages/react/services/filestore.service.ts
+++ b/packages/react/services/filestore.service.ts
@@ -5,6 +5,7 @@ type UploadFileProps = {
   scope: string;
   directory: string;
   name: string;
+  disableLoading?: boolean;
   onProgress?: (progress: number) => void;
 };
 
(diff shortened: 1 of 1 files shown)


packages/react/services/filestore.service.ts
```diff
@@ -21,24 +22,26 @@ const initFilestoreService = (axios: AxiosInterceptor) => {
     scope,
     directory,
     name,
+    disableLoading,
     onProgress,
   }: UploadFileProps) => {
     const { data } = await axios.get(`${apiBaseUrl}/upload`, {
       requestKey: `filestore/presign/file/${name}`,
-      isInterruptive: true,
+      isInterruptive: disableLoading? false : true,
```
wengkhing, line 30:
```suggestion
      isInterruptive: !disableLoading
```

wengkhing (approved):
Just nitpick else looks good

PR #72, opened 2025-04-27
fix axios interceptor removing all headers

jaydenpung:
Currently if you do something like 
```
axios.put(data.url, file, {
      requestKey: `filestore/upload/${name}`,
      isInterruptive: !disableLoading,
      feedback: {
        loading: 'Uploading file',
      },
      headers: { // added this
        'Content-Type': file.type,
        'Content-Disposition': 'inline', 
      },
      onUploadProgress: (progressEvent) => {
        if (onProgress && progressEvent.total) {
          const progress = Math.round(
            (progressEvent.loaded / progressEvent.total) * 100,
          );
          onProgress(progress);
        }
      },
    });
    ```
    The entire headers will be removed, this PR put it back in

### DIFF
--- a/packages/react/services/filestore.service.ts
+++ b/packages/react/services/filestore.service.ts
@@ -35,7 +35,6 @@ const initFilestoreService = (axios: AxiosInterceptor) => {
         scope,
         directory,
         filename: name,
-        fileType: file.type,
       },
     });
 
@@ -47,6 +46,7 @@ const initFilestoreService = (axios: AxiosInterceptor) => {
       },
       headers: {
         'Content-Type': file.type,
+        'Content-Disposition': 'inline',
       },
       onUploadProgress: (progressEvent) => {
         if (onProgress && progressEvent.total) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/react/services/filestore.service.ts [hunk 1]
```diff
@@ -35,7 +35,6 @@ const initFilestoreService = (axios: AxiosInterceptor) => {
         scope,
         directory,
         filename: name,
-        fileType: file.type,
```
jaydenpung, line 38: no longer required at presign

packages/react/services/filestore.service.ts
```diff
@@ -47,6 +46,7 @@ const initFilestoreService = (axios: AxiosInterceptor) => {
       },
       headers: {
         'Content-Type': file.type,
+        'Content-Disposition': 'inline',
```
jaydenpung, line 49: maek files viewable instead of download on open. This works for file types that is suitable for view in browser otherwise they would still be downloaded. So I think its safe to put this as default

jyungtong (approved):

PR #76, opened 2025-04-29
Feat/unique field implementation

wengkhing:
## What

Introduction usage of `uniqueFields`.

### Behaviour

- If `uniqueFields` is specified, entity creation and updates will include a check to ensure that all values of the listed fields are unique within the scope of that entity type.
- If value already exists, creation/updates would throw validation error, eg. `username 'johndoe' already exists`.
- Unique value can also serve as a entry point to retrieve entity, `getEntityByUniqueField(Entity.User, 'username', 'johndoe')`
- On entity deletion, any associated unique field values will be released, making them available for reuse.

## Why

Scalable and flexible way of handling unique values for an entity type such as email, username, identity number etc.

### DIFF
--- a/packages/core/data/Entity.ts
+++ b/packages/core/data/Entity.ts
@@ -268,10 +268,60 @@ export class EntityRepository extends Repository {
     return;
   }
 
+  async getEntityByUniqueField<T extends EntityType>(
+    entityType: T,
+    fieldName: string,
+    value: string,
+  ): Promise<Entity<T>> {
+    const resp = await this.dynamodbClient.query({
+      TableName: this.TABLE_NAME,
+      KeyConditionExpression: '#PK = :PK and begins_with(#SK, :SK)',
+      ExpressionAttributeNames: {
+        '#PK': 'PK',
+        '#SK': 'SK',
+      },
+      ExpressionAttributeValues: {
+        ':PK': { S: `UNIQUE#${fieldName}#${value}` },
+        ':SK': { S: entityType },
+      },
+    });
+
+    return Entity.fromItem(resp.Items?.[0]);
+  }
+
+  async getUniqueFieldValueAvailability<T extends EntityType>(
+    entityType: T,
+    fieldName: string,
+    value: string,
+  ): Promise<void> {
+    const resp = await this.dynamodbClient.query({
+      TableName: this.TABLE_NAME,
+      KeyConditionExpression: '#PK = :PK and begins_with(#SK, :SK)',
+      ExpressionAttributeNames: {
+        '#PK': 'PK',
+        '#SK': 'SK',
+      },
+      ExpressionAttributeValues: {
+        ':PK': { S: `UNIQUE#${fieldName}#${value}` },
+        ':SK': { S: entityType },
+      },
+    });
+
+    if (resp.Items?.[0]) {
+      throw new StandardError(
+        'UNIQUE_VALUE_EXISTS',
+        `${fieldName} '${value}' already exists`,
+      );
+    }
+
+    return;
+  }
+
   createEntityTransactItems<T extends EntityType>(
     entity: Entity<T>,
-    opts?: {
+    opts: {
       mutualId?: string;
+      uniqueFieldValues: Record<string, string>;
     },
   ): TransactWriteItem[] {
     const TransactItems: TransactWriteItem[] = [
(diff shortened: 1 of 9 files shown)


packages/core/data/__tests__/Entity.test.ts
```diff
@@ -269,18 +272,25 @@ describe('Entity & EntityRepository', () => {
   describe('EntityRepository', () => {
     // Use MockEntityType enum
     let createdUser: Entity<EntityType>;
+    let updatedUser: Entity<EntityType>;
     const userEmail = `test-${ulid()}@example.com`;
+    const username = `test-${ulid()}`;
```
jyungtong, line 277: Can we have a test to ensure recreating with the same `username` will failed.

Based on what I found, the current tests only check with the `getUniqueFieldValueAvailability()`. When race condition happened and `getUniqueFieldValueAvailability()` passed in both executions, we want only 1 get through it and another execution fail.

packages/core/data/Entity.ts
```diff
@@ -268,10 +268,60 @@ export class EntityRepository extends Repository {
...
+      ExpressionAttributeValues: {
+        ':PK': { S: `UNIQUE#${fieldName}#${value}` },
+        ':SK': { S: entityType },
+      },
+    });
+
+    return Entity.fromItem(resp.Items?.[0]);
+  }
+
+  async getUniqueFieldValueAvailability<T extends EntityType>(
```
jyungtong, line 292: i think this is more like a business specific logic. 

Why can't the consumer app use `getEntityByUniqueField()` and decide on their own?

packages/core/data/Entity.ts
```diff
@@ -342,8 +410,47 @@ export class EntityRepository extends Repository {
...
+          );
+        }
+        return this.getUniqueFieldValueAvailability(
+          entityType,
+          field,
+          (entityPayload as Record<string, string>)[field],
+        );
+      });
+
+      await Promise.all(uniqueFieldAvailabilityPromises);
```
jyungtong, line 440: can we delegate the unique availability check to consumer app instead of data layer? 

I believe the transaction should already do it's job to prevent race condition.

packages/core/data/__tests__/Entity.test.ts
wengkhing: https://github.com/monorist/monorise/pull/76/commits/703989d30af5258c055f492480b7a74890a2cfc2

packages/core/data/Entity.ts
wengkhing: Similar to what we had implemented earlier for [email availability check](https://github.com/monorist/monorise/pull/76/files#diff-7ecfab45e52c101354f8fdce2519496f35a30a0e1fbff7b44569b5a9370a2ef7R247). I think would be a useful utility as a framework, any way, business would still have to implement it themselves, why not prepare a reusable utility function?

packages/core/data/Entity.ts
```diff
@@ -417,6 +578,72 @@ export class EntityRepository extends Repository {
...
+            );
+          }
+          return this.getUniqueFieldValueAvailability(
+            entityType,
+            field,
+            (toUpdate.data as Record<string, string>)[field],
+          );
+        });
+
+        await Promise.all(uniqueFieldAvailabilityPromises);
```
jyungtong, line 631: same to `createEntity()`'s comment

packages/core/data/__tests__/Entity.test.ts
```diff
@@ -312,6 +319,17 @@ describe('Entity & EntityRepository', () => {
       ).rejects.toThrow(); // Should throw due to ConditionExpression failure
     });
 
+    it('should fail to create an entity with existing unique field value', async () => {
+      await expect(
+        // Use MockEntityType enum
+        entityRepository.createEntity(
+          MockEntityType.USER as unknown as EntityType,
+          { name: 'Duplicate', username: createdUser.data.username },
```
jyungtong, line 327: need to update the available field in the mock https://github.com/wengkhing/monorise/blob/703989d30af5258c055f492480b7a74890a2cfc2/packages/core/data/__tests__/Entity.test.ts#L5

jyungtong (commented):
after having conversation, we are aligned to have service layer and service layer test later, as for now, it's not best structured but at least it's tested.

jyungtong (approved):

PR #78, opened 2025-04-29
implement useEntityByUniqueField hook

wengkhing:
## What

Implement `useEntityByUniqueField` for @monorise/react with cache control


packages/core/controllers/setupRoutes.ts
```diff
@@ -49,7 +49,7 @@ const setupCommonRoutes =
       container.createEntityController.controller,
     );
     router.get(
-      '/entity/:entityType/:byUniqueField/:byUniqueFieldValue',
+      '/entity/:entityType/unique/:byUniqueField/:byUniqueFieldValue',
```
jovanimal, line 52: @wengkhing does the word `by` actually trying to implying anything than it otherwise couldn't? just wondering if this endpoint will be good enough if it were to be interpreted as `/entity/:entityType/unique/:uniqueField/:uniqueFieldValue`?

packages/react/services/core.service.ts
```diff
@@ -121,6 +121,27 @@ const initCoreService = (
...
+    fieldName: string,
+    value: string,
+    opts: CommonOptions = {},
+  ) => {
+    const { entityApiBaseUrl = ENTITY_API_BASE_URL } = options;
+    return axios.get<CreatedEntity<T>>(
+      opts.customUrl ||
+        `${entityApiBaseUrl}/${entityType}/unique/${fieldName}/${value}`,
+      {
+        requestKey: `entity/${entityType}/unique/${fieldName}/${value}`,
```
jovanimal, line 135: I remember @yeuharn96 implemented a helper for requestKey, shall we implement a helper function like `uniqueStateKey`? similar to the one [here](https://github.com/monorist/monorise/pull/62/files#diff-a06deb491d3f9f14813da79af4cf6ded325191d93daa043d701bf5f6a8c6ec43R26)

packages/react/actions/core.action.ts
```diff
@@ -798,6 +836,52 @@ const initCoreActions = (
...
+    isFirstFetched?: boolean;
+    refetch: () => Promise<CreatedEntity<T> | undefined>;
+  } => {
+    const dataMap = monoriseStore(
+      (state) => state.entity[entityType]?.dataMap || new Map(),
+    );
+    const isFirstFetched = monoriseStore(
+      (state) => state.entity[entityType]?.isFirstFetched,
+    );
+    const requestKey = `entity/${entityType}/unique/${field}/${value}`;
```
jovanimal, line 858: can then use the `uniqueStateKey` here too

packages/react/services/core.service.ts
```diff
@@ -121,6 +121,27 @@ const initCoreService = (
     );
   };
 
+  const getEntityByUniqueField = <T extends Entity>(
+    entityType: T,
+    fieldName: string,
```
jovanimal, line 126: i noticed in the params we use `fieldName` and `field` interchangeably, will it be better if we make the naming usage consistent?

packages/core/controllers/setupRoutes.ts
wengkhing: Sounds good to me, will rename the path parameters accordingly

packages/react/lib/utils.ts
```diff
@@ -30,3 +30,7 @@ export const tagStateKey = (
 ) => {
   return `${entityType}/${tagName}/${group || ''}`;
 };
+
+export const uniqueFieldStateKey = (fieldName: string, fieldValue: string) => {
```
jovanimal, line 34: on second thought, do you think it will be better if all these 3 helpers are prepend with the verb `get`? i.e. `getMutualStateKey`, `getTagStateKey`, `getUniqueFieldStateKey`.

It looks more intuitive and imply that it's a helper function where it's being consumed,  e.g.
```
const stateKey = getUniqueFieldStateKey(fieldName, value);
```
instead of 
```
const stateKey = uniqueFieldStateKey(fieldName, value) 
```

packages/react/actions/core.action.ts
```diff
@@ -798,6 +842,53 @@ const initCoreActions = (
     };
   };
 
+  const useEntityByUniqueField = <T extends Entity>(
+    entityType: T,
+    fieldName: string,
+    value?: string,
```
jyungtong, line 848: any reason this is optional? as i noticed, other places are mandatory. eg https://github.com/wengkhing/monorise/blob/2f853d108f7884bbe5cfd1becc31e069d2cf4afe/packages/react/actions/core.action.ts#L233

packages/react/services/core.service.ts
jyungtong: is this resolved on code wise? didn't see the change

jyungtong:
is this missed https://github.com/monorist/monorise/pull/78#discussion_r2068164429 ?

packages/react/lib/utils.ts
wengkhing: Make sense, should always use verb for function name, I will make an adjustment

packages/react/actions/core.action.ts
wengkhing, line 848: @jyungtong 
The idea behind this is similar to `useEntity()` where the `entityId` in optional. The reason is that in front-end, sometimes the hook might have dependencies on other Promise values. Just a rough example for you to get an idea, imagine in the same component:
```
const user = useEntity(Entity.User, profileId);
const profile = useEntityByUniqueField(Entity.Profile, 'email', user.data.email);
```
So in the example, `useEntityByUniqueField` `value` might be empty on first load. I foresee scenario like this might happen therefore it is implemented this way.

packages/react/services/core.service.ts
wengkhing: https://github.com/monorist/monorise/pull/78/commits/4715fb980e63cbaf136af181c11549e942d3cede

jyungtong (approved):
LGTM!

jovanimal (approved):

PR #99, opened 2025-05-28
feat(cli): add dev mode

wengkhing:
## What 

- Add cli to enable dev mode
- removed `monorise` command, instead broken down into 2 different commands

Currently there are 2 different commands:
- `monorise dev`: Generate monorise index file, and start `sst dev`, at the same time also watch changes in the monorise config directory
- `monorise build`: Just generate monorise index file

## Why

Improve developer experience. Developer doesn't need to know when to run the `monorise` script again. Any changes in entity config would automatically updated during local dev mode.

## How to implement

In your `package.json`,  instead of this
```
{
  // ...
  "scripts": {
    "dev": "sst dev"
    "build": "sst build"
  }
}
```
now you could do this:
```
{
  // ...
  "scripts": {
    "dev": "monorise dev"
    "build": "monorise build"
  }
}
```

### DIFF
--- a/.changeset/olive-trees-create.md
+++ b/.changeset/olive-trees-create.md
@@ -1,0 +1,5 @@
+---
+"@monorise/cli": patch
+---
+
+add cli dev mode
--- a/.changeset/olive-trees-create.md
+++ b/.changeset/olive-trees-create.md
@@ -1,5 +1,5 @@
 ---
-"@monorise/cli": patch
+"@monorise/cli": minor
 ---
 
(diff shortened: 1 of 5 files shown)


.changeset/olive-trees-create.md
```diff
@@ -0,0 +1,5 @@
+---
+"@monorise/cli": patch
```
jyungtong, line 2: wdyt a `minor`?

yeuharn96 (approved):

PR #106, opened 2025-06-05
list entities and mutuals accepts any string as projection expression

wengkhing:


jyungtong (approved):
Lgtm, just curious on your use case

.changeset/forty-rats-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@monorise/core": patch
+---
+
+no data projection expression should return dates
```
jovanimal, line 5: cmiiw, since the description here seems different from the code changes.
Do you mean both the projection expression `NO_DATA` and `MUTUAL_DATA_ONLY` also should return dates, instead of just `NO_DATA` only?

wengkhing: Sorry for the confusing comment, it's both, I think it would make sense for both to return dates.

jyungtong (approved):

PR #108, opened 2025-06-06
Upgrade zod to v4

wengkhing:
## What

- Bump zod to v4 (major version)
- Simplifies `createEntityConfig`, we could remove `effects` field entirely
- All schemas, `baseSchema`, `createSchema` and `mutualSchema` supports refine.

## Why
- Major performance upgrade especially when resolving types, [almost x100 faster
](https://zod.dev/v4?id=100x-reduction-in-tsc-instantiations)
## Note
Zod 3.25.0 onwards include v4 codes, there isn't an actual v4 package yet but this could help [transition to the actual v4](https://zod.dev/v4?id=versioning) package easily

## How to use

With `effects` gone, how can we refine schema? Here's a simple example, simple enough?

```js
import { createEntityConfig } from '@monorise/base';
import { z } from 'zod/v4';

const baseSchema = z
  .object({
    name: z.string(),
    age: z.number(),
  })
  .refine(
    (value) => {
      if (value.age < 16) {
        return false;
      }

      return true;
    },
    {
      message:
        'Must be at least 16 years old',
      path: ['options'],
    },
  );

const config = createEntityConfig({
  name: 'user',
  displayName: 'User',
  baseSchema,
  searchableFields: ['name'],
});

export default config;
```

### DIFF
--- a/.changeset/legal-areas-sell.md
+++ b/.changeset/legal-areas-sell.md
@@ -1,0 +1,5 @@
+---
+"@monorise/base": minor
+---
+
+Upgrade zod to v4, retire `effects` field since we can merge refined shapes, improves usability of `createEntityConfig`
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -2,6 +2,7 @@
   "name": "@monorise/base",
   "version": "0.0.4",
   "description": "",
+  "type": "module",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "directories": {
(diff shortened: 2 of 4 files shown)


jyungtong:
is there any impact on existing implementation in our consumer app?

Do we need to change all import to 'zod/v4' to use this?

wengkhing:
> is there any impact on existing implementation in our consumer app?
> 
> Do we need to change all import to 'zod/v4' to use this?

Yes, it's a breaking change unfortunately. However, still haven't sort out typing issue for react-hook-form. The form seems to work without issue, but the type doesn't resolve correctly, but wow just in time, the resolved for zod v4 just released 2 days ago: https://github.com/react-hook-form/resolvers/issues/768

PR #111, opened 2025-06-12
chore: standardized error handling in promise call

jyungtong:
# why
- a lot unhandled error throw into our error tracking service, which could consume $$$

### DIFF
--- a/packages/react/actions/core.action.ts
+++ b/packages/react/actions/core.action.ts
@@ -38,6 +38,13 @@ import type { AppActions } from './app.action';
 // USE store.getState() WHEN using within an action such as getEntity, getProfile
 // USE store() WHEN using within a react hook so that it gets the benefit of reactivity, such as useProfile, useEntity
 
+const defaultOnError = (error: ApplicationRequestError | Error) => {
+  if (!(error instanceof Error) && error.status >= 500) {
+    // we only want server error to be captured by error tracking service
+    throw error;
+  }
+};
+
 const initCoreActions = (
   monoriseStore: MonoriseStore,
   appActions: AppActions,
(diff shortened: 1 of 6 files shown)


packages/react/actions/core.action.ts
```diff
@@ -404,37 +404,44 @@ const initCoreActions = (
...
-          dataMap: convertToMap(entities, 'entityId'),
-          isFirstFetched: true,
-          lastKey,
-        };
-      }),
-      undefined,
-      `mr/mutual/list/${selfKey}`,
-    );
+      return { error, data: null };
+    }
```
wengkhing, line 444: Is this necessary? You could still return the `error` from line 403, isn't it?

packages/react/actions/core.action.ts
```diff
@@ -404,37 +404,44 @@ const initCoreActions = (
...
+      return { error };
     }
 
-    const { data } = await mutualService.listEntitiesByEntity(
-      id,
-      opts,
-      chainEntityQuery,
-    );
-    const { entities, lastKey } = data;
+    try {
```
wengkhing, line 410: If we are doing `try...catch` here, should we be doing the same for other actions, else it's going to be inconsistent with other actions.

wengkhing:
> a lot unhandled error throw into our error tracking service, which could consume $$$

could this be selectively silenced from the error tracking service?

packages/react/actions/core.action.ts
jyungtong: I believe this return is for immediate return, if we rely on L403, there could be situation where it is still `null / undefined`.

jyungtong:
after discussion with @wengkhing , i will make a POC of app level higher order hook on top of useMutuals and see if it's feasible and practical.

packages/react/actions/core.action.ts
```diff
@@ -221,22 +284,34 @@ const initCoreActions = (
     const isLoading = checkIsLoading(requestKey);
     const error = getError(requestKey);
     const { forceFetch } = opts;
+    const onError = opts.onError ?? defaultOnError;
 
     if (!forceFetch && (entity || isLoading || error)) {
-      return entity;
+      if (error) {
```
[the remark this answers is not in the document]
jyungtong: https://github.com/monorist/monorise/pull/111/commits/8a026f618541d906f1ed5dad06167878598f82b5

jovanimal:
oppps comment in the commit itself doesn't show up here. 
commented [here](https://github.com/monorist/monorise/commit/8a026f618541d906f1ed5dad06167878598f82b5#diff-6b48af6760bb32105adb00c4269945b21fda4daf80b5d0bdcf60c2f29f00b53bR291)

PR #113, opened 2025-06-13
Chore/restructure folder

jyungtong:
# problem
- when running `npm run dev`,  @monorise/core always has an issue conflicting with the folder

# solution
- isolate main logic into `src` folder respectively

# piggyback
- use `@tsconfig/node22`
- extend all sub-repos with `tsconfig.base.json`

# out of scope
- no business logic change


jyungtong:
too complex, we can just ignore `dist` folder in tsconfig.json

jyungtong:
too complex, we can just ignore dist folder in tsconfig.json

PR #117, opened 2025-07-12
Monorise core component

wengkhing:
## What

- Upgrade monorise to adapt SST v3
- Create a super component `monorise.module.core`

## How to use?

By defining the following, the entire monorise stack including all the processors are automatically created.
```
// add the following in sst.config.ts run function
    const { monorise } = await import("@monorise/sst");
    const { bus, api, alarmTopic } = new monorise.module.Core('test-app', {
      allowOrigins: ['http://localhost:3000'],
    });
// infra resources such as infra event bus, api gateway can also be retrieved from monorise core super component
```

## Why

This would allow developers quickly create an entire scalable, maintainable, cost-optimized, secured backend code, without having to write boilerplate codes

## What's Next?

Documentation for human reference and also LLM for future project generation


src/react/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  useTaggedEntities,
+  useEntityState,
+  getEntityRequestKey,
+  getMutualRequestKey,
+  getTagRequestKey,
+  getUniqueFieldRequestKey,
+  getEntity,
+} from '../../packages/react';
+
+export default Monorise;
```
jyungtong, line 56: nit: should place at last of the LOC

packages/cli/commands/build.ts
```diff
@@ -0,0 +1,9 @@
+import { spawn } from 'node:child_process';
```
jyungtong, line 1: not use

packages/cli/commands/dev.ts
```diff
@@ -0,0 +1,64 @@
+import { spawn } from 'node:child_process';
```
jyungtong, line 1: not use

packages/cli/cli.ts
```diff
@@ -2,9 +2,9 @@
 
 import 'tsx';
 import 'tsconfig-paths/register.js';
+import { spawn } from 'node:child_process';
```
jyungtong, line 5: not use

packages/sst/.sst/provider-lock.json
jyungtong, line 1: don't think we need this file and folder based on the gitignore

packages/core/controllers/setupRoutes.ts
```diff
@@ -1,81 +1,71 @@
-import type { Router } from 'express';
+import type { Entity as EntityType, createEntityConfig } from '@monorise/base';
```
jyungtong, line 1: not use

packages/sst/.sst/provider-lock.json
wengkhing: I've purposely let this file slide into commit. We will [need this file](https://github.com/monorist/monorise/pull/117/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R75) else [build wouldn't work](https://github.com/monorist/monorise/actions/runs/16236847198/job/45848113614).

PR #120, opened 2025-07-25
feat: support limit mutuals returned

jyungtong:
https://github.com/user-attachments/assets/784d8da0-c6c8-4276-add1-8e583706d025

### DIFF
--- a/packages/react/actions/core.action.ts
+++ b/packages/react/actions/core.action.ts
@@ -1367,7 +1367,7 @@ const initCoreActions = (
     error?: ApplicationRequestError;
     isFirstFetched?: boolean;
     lastKey?: string;
-    listMore: () => void;
+    listMore?: () => void;
   } => {
     const stateKey = getMutualStateKey(
       byEntityType,
--- a/packages/react/actions/core.action.ts
+++ b/packages/react/actions/core.action.ts
@@ -591,11 +591,25 @@ const initCoreActions = (
           for (const [key, value] of newEntityDataMap) {
             state.entity[entityType]?.dataMap.set(key, value);
           }
-          state.mutual[selfKey] = {
-            dataMap: convertToMap(entities, 'entityId'),
-            isFirstFetched: true,
-            lastKey,
-          };
+
+          // prepare or extend the mutual entry
+          const existing = state.mutual[selfKey];
+          if (existing) {
+            // extend the existing map
+            const incoming = convertToMap(entities, 'entityId');
+            for (const [k, v] of incoming) {
+              existing.dataMap.set(k, v);
+            }
+            existing.isFirstFetched = true;
+            existing.lastKey = lastKey;
+          } else {
+            // first time: create it
+            state.mutual[selfKey] = {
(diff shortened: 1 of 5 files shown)


jyungtong:
@yeuharn96 @wengkhing have you already using `listMore()` returned from `useMutuals()`?

jovanimal:
@jyungtong the UI of after clicking `Show more` looks different to the existing way of fetching more entities, example [this page](https://dev-manage.gift-ed.com/learner). Where we click on `Show more` and the next X items show up and added to the existing list, instead of the recording above, it refreshed the whole list.

Is this something to do with our app or the way of fetching?

jyungtong:
> @jyungtong the UI of after clicking `Show more` looks different to the existing way of fetching more entities, example [this page](https://dev-manage.gift-ed.com/learner). Where we click on `Show more` and the next X items show up and added to the existing list, instead of the recording above, it refreshed the whole list.
> 
> Is this something to do with our app or the way of fetching?

for UI (gift-ed) related, let's bring it back to our UI PR, i believe `monorise` is to support the limit feature, however, UI related logic should be the consumer responsibility.

putting the demo here is to showcase how it will looks like.

jovanimal:
@jyungtong yes i am aware of the segregation between `monorise` and `gift.ed`. That was the reason I wanted to confirm the below earlier, whether it's our `gift.ed` app, _or_ the way of fetching data in `monorise`. 

> Is this something to do with our app or the way of fetching?

Looks like it's the our `gift.ed` responsibility here, I have my answer.

packages/react/actions/core.action.ts
```diff
@@ -1421,22 +1435,24 @@ const initCoreActions = (
...
-              ...opts,
-              forceFetch: true,
-              params: { ...opts.params, lastKey },
-              stateKey,
-            },
-            chainEntityQuery,
-          );
-        }
-      },
+      ...(lastKey && {
```
jovanimal, line 1438: what happens if we dont check for the existence of `lastKey` before appending  `listMore`? 
asking for consistency purpose with `useTaggedEntities` [here](https://github.com/monorist/monorise/pull/120/files#diff-6b48af6760bb32105adb00c4269945b21fda4daf80b5d0bdcf60c2f29f00b53bR1501-R1518).

jovanimal (approved):

packages/react/actions/core.action.ts
jyungtong: in previous version, the `listMore` is always returned whether or not the mutual has reached the end. The consumer app has to figure out if there's `lastKey` before to decide to use `listMore`. I think it is defeating the purpose `listMore` when there's nothing more to actually `listMore`.

@wengkhing @yeuharn96 what's your thought on this design, if we are to only return `listMore` when it's sure there's more to list.

jovanimal: fair point, semantically it makes more sense to have `listMore` only when there is more and not the end yet. 
sgtm to check for `lastKey`.

PR #140, opened 2025-08-29
fix: unique field replication

yeuharn96:
- potential bug: if a unique field record (eg: `UNQUE#email#xxx`) is inserted without `updatedAt` attribute, replication processor will not recognize it in subsequent entity updates
- to reproduce:
  - find a unique field record `UNQUE#email#xxx` in dynamodb
  - remove the `updatedAt` attribute
  - perform update on its corresponding `#METADATA#` entity
  - the unique field record will not get update with latest data, because the `ConditionExpression: #${targetUpdatedAt} < :${targetUpdatedAt}` is always false when `updatedAt` attribute is not exists

### DIFF
--- a/.changeset/sour-mirrors-care.md
+++ b/.changeset/sour-mirrors-care.md
@@ -1,0 +1,5 @@
+---
+"@monorise/core": patch
+---
+
+bug fix on unique field record will not get update if `updatedAt` attribute is not exists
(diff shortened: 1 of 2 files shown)


jyungtong:
possible to add the `updatedAt` field to unique record instead?

the replicator has been working fine for quite some time, I'm afraid if we change the infra layer  might cause more implications.

jyungtong:
If you believe this should works, can you add tests which is related to the change in unique data?

yeuharn96:
ok, will try another approach

PR #143, opened 2025-09-04
Feat/access dependency container custom route

jyungtong:
# what

- allow custom routes to access dependency container instance so that it can publish event.


packages/core/handles/app.ts
```diff
@@ -8,33 +8,37 @@ import generalErrorHandler from '../middlewares/general-error-handler';
...
-    if (routes) {
-      app.route('/app', routes);
-    }
+      if (routes) {
+        if (typeof routes !== 'function') {
+          app.route('/app', routes);
+        } else {
+          app.route('/app', routes(container));
+        }
+      }
```
jyungtong, line 37: main changes, the rest are formatting

PR #145, opened 2025-09-15
feat: combine monorise package into one

wengkhing:
## What

This change combine all monorise packages into one single package `monorise`, so basically user would just have to:
`npm install monorise`
instead of:
`npm install @monorise/react @monorise/base @monorise/core @monorise/sst @monorise/cli`.

With this change, it is also possible to run `npx monorise init` without first installing `monorise` locally.

- Also added a cool ASCII logo of monorise when running `monorise init` or `monorise dev`

## Why

- Improve developer experience not having to install multiple packages.
- Improve branding =)

### DIFF
--- a/packages/cli/cli.ts
+++ b/packages/cli/cli.ts
@@ -223,7 +223,7 @@ async function generateHandleFile(
   // If customRoutesPath is not provided, routesImportLine remains empty and appHandlerPayload remains `{}`
 
   const combinedContent = `
-import CoreFactory from '@monorise/core';
+import CoreFactory from 'monorise/core';
 import config from './config';
 ${routesImportLine ? `${routesImportLine}\n` : ''}const coreFactory = new CoreFactory(config);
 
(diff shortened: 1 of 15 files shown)


jyungtong (commented):
the `prompt-doc.md` is empty, is it expected?

packages/monorise/build.js
jyungtong, line 1: shall we gitignore this file?

packages/monorise/index.ts
jyungtong, line 1: can we put this file into `src` folder? so the ts build can focus on the particular folder

packages/cli/cli.ts
```diff
@@ -269,8 +269,113 @@ async function generateFiles(): Promise<string> {
...
+      lines.splice(configEndIndex, 0, ...turbopackConfig);
+      return lines.join('\n');
+    }
+  }
+
+  // If we can't parse it, return original content
+  return configContent;
+}
+
+const MONORISE_LOGO = `
```
jyungtong, line 327: cantek

packages/cli/cli.ts
```diff
@@ -359,10 +464,40 @@ export default config;
...
+  if (fs.existsSync(nextConfigPath)) {
+    try {
+      const nextConfigContent = fs.readFileSync(nextConfigPath, 'utf8');
+
+      // Check if turbopack config already exists
+      if (
+        !nextConfigContent.includes('turbopack') &&
+        !nextConfigContent.includes('resolveAlias')
+      ) {
+        const updatedConfig = addTurbopackConfig(nextConfigContent);
```
jyungtong, line 478: does that mean monorise will have dependencies on turbopack?

wengkhing: Addressed no longer relevant

PR #200, opened 2026-03-26
feat: add cost optimization support and documentation

wengkhing:
## Summary

Adds cost optimization support to Monorise SST components and documentation.

## SST Components Changes

- SingleTable: Add billing mode, capacity config, auto-scaling
- QFunction: Export types and expose full Function configuration
- MonoriseCore: Add apiConfig, processorConfig, replicatorConfig
- Export new types: BillingMode, CapacityConfig, FunctionConfig

## Documentation Changes

- New Cost Optimization guide with 3-phase strategy
- FAQ updates for serverless cost questions
- Added to sidebar navigation

## Changeset

In dev mode for prerelease versioning.


jovanimal:
@wengkhing curious, what's the trigger for you to initiate this? can share a bit of background story? 

is this urgent, would like to go through this but probably can do so only later today

package.json
```diff
@@ -31,6 +31,7 @@
   "devDependencies": {
     "@biomejs/biome": "^1.9.4",
     "@changesets/cli": "^2.28.1",
+    "@fission-ai/openspec": "^1.2.0",
```
jyungtong: have we started using this?

wengkhing: Removed — confirmed unused, dropped the `@fission-ai/openspec` devDependency and regenerated the lockfile.

wengkhing:
Rebased onto latest main (this branch was 73 commits behind — predates the whole PR #234/#235/#240/#203 wave) and resolved conflicts:

- **`packages/sst/components/single-table.ts`**: merged main's `fromTableName` (link to an existing table) with this PR's `billingMode`/capacity/auto-scaling additions. Kept `ttl: 'expiresAt'` hardcoded per main's current convention (main had already removed the configurable `ttl?: string` this PR's diff was based on — TTL attribute name is always `expiresAt` by framework design). Skip auto-scaling setup when `fromTableName` is used (capacity is managed wherever that table is actually defined).
- **`packages/sst/components/q-function.ts`**: merged cleanly, but caught a real bug in this PR's own diff while resolving — `maxBatchingWindow?: Input<DurationMinutes>` uses a bare `Input` type that doesn't exist as a global here (checked — nothing in this codebase exposes one). Fixed to `$util.Input<DurationMinutes>`, matching the pattern already used in `single-table.ts`/`monorise-core.ts`.
- **`packages/sst/components/monorise-core.ts`**: merged main's `fromTableName`/`cloudwatchLogRetention`/`cloudwatchDashboard` toggle with this PR's `apiConfig`/`processorConfig`/`replicatorConfig`/table billing options — all independent, non-overlapping additions.
- **`.changeset/pre.json`**: this PR's branch predates the current release process and reintroduced a stale `pre.json` (prerelease "pre" mode) that doesn't exist on main today. Dropped it — merging it would've put the whole repo back into dev-tag prerelease mode.
- **`package-lock.json`**: regenerated via `npm install` rather than hand-resolving.

Verified the SST changes type-check cleanly: `tsc --noEmit` against `index.ts` + `.sst/platform/config.d.ts` (the ambient SST globals) reports zero errors.

Also, per this thread:
- Removed the unused `@fission-ai/openspec` devDependency (regenerated lockfile).
- Added `monorise` to the cost-optimization changeset — same gap as PR #203: the combined package bundles `@monorise/sst`'s compiled output via `monorise/sst` but has no npm dependency link to it, so changesets' automatic cascade wouldn't have caught this.

PR #201, opened 2026-03-26
Websocket #1 - feat(websocket): add WebSocket layer for real-time updates

wengkhing:
## Summary

Add optional WebSocket support to monorise for real-time entity updates.

**Philosophy:** HTTP for mutations (reliable), WebSocket for real-time updates (scalable)

## Key Features

- **Initial fetch via HTTP** — Reliable, with pagination
- **Real-time updates via WebSocket** — Live sync when data changes
- **Auto-refetch on reconnect** — Automatically catches up after disconnect

### SST Infrastructure
- Add `webSocket` config option to MonoriseCore
- Lambda handlers: `$connect`, `$disconnect`, `$default`, `broadcast`
- DynamoDB Streams integration for change broadcasting
- Connection/subscription tracking in shared table using `#` separators and `#METADATA#` SK (aligned with entity schema)
- Handler paths use `.monorise/handle.ts` pattern (CoreFactory)

### Core
- WebSocket handlers moved to `packages/core/processors/websocket-processor.ts`
- Unmarshalls DynamoDB stream images before broadcasting
- `ConsistentRead: true` on all broadcast subscriber queries

#### `useEntitySocket(entityType, opts?)`
Fetch initial data via HTTP, subscribe to real-time updates via WebSocket. Returns arrays (consistent with `useEntities`).

#### `useMutualSocket(byEntityType, byEntityId, mutualEntityType, opts?)`
Same pattern for mutual relationships. Returns arrays (consistent with `useMutuals`).

#### `useEphemeralSocket(channel, opts?)`
Non-persisted real-time messages (typing indicators, presence).

### New Package
- `monorise/proxy` — Server-side helpers for proxy routes (e.g., `generateWebSocketTicket()`)

## Usage

```typescript
// 1. Enable in SST config
const core = new monorise.module.Core('app', {
  webSocket: { enabled: true }
})

// 2. Use hooks — same API as useEntities/useMutuals
const { entities } = useEntitySocket(Entity.CHANNEL)
const { mutuals } = useMutualSocket(Entity.CHANNEL, channelId, Entity.MESSAGE)
```

## Follow-up PRs

- #215 — Ticket system for entity feed auth
- #216 — Feed subscription broadcast resolution
- #217 — `useEntityFeed` hook
- #218 — Connection cleanup via R1 GSI
- #219 — Chat example with entity feed

## Test plan

- [ ] Enable WebSocket in SST config, verify API Gateway WebSocket is created
- [ ] Connect via WebSocket, verify connection record in DynamoDB
- [ ] Subscribe to entity type, verify entity changes are broadcasted
- [ ] Subscribe to mutual type, verify mutual changes are broadcasted
- [ ] Test ephemeral messages between two connections
- [ ] Test reconnection and auto-refetch


packages/cli/cli.ts
```diff
@@ -229,6 +229,7 @@ async function generateHandleFile(
   // Detect whether the consumer uses the combined 'monorise' package or scoped '@monorise/*' packages
   const usesCombinedPackage = fs.existsSync(path.join(projectRoot, 'node_modules', 'monorise'));
   const coreImportPath = usesCombinedPackage ? 'monorise/core' : '@monorise/core';
+  const sstImportPath = usesCombinedPackage ? 'monorise/sst' : '@monorise/sst';
```
jyungtong: not used?

packages/sst/index.ts
```diff
@@ -9,3 +9,6 @@ export const monorise = {
     QFunction,
   },
 };
+
+
+
```
jyungtong: empty space?

packages/sst/components/monorise-core.ts
```diff
@@ -149,6 +161,80 @@ export class MonoriseCore {
...
+
+      // Subscribe broadcast handler to DynamoDB Stream
+      this.table.table.subscribe(
+        `${id}-ws-broadcast`,
+        {
+          name: `${$app.stage}-${$app.name}-${id}-ws-broadcast`,
+          handler: `${dotMonorisePath}/handle.wsBroadcast`,
+          runtime,
+          memory: '1024 MB',
+          timeout: '60 seconds',
```
jyungtong: not using args?

packages/sst/handlers/websocket.ts
```diff
@@ -0,0 +1,569 @@
...
+    | 'pong';
+  id: string;
+  payload: unknown;
+}
+
+const getTableName = () => process.env.CORE_TABLE || '';
+
+const getWsEndpoint = () => {
+  const apiId = process.env.WEBSOCKET_API_ID;
+  const region = process.env.AWS_REGION || 'us-east-1';
```
jyungtong: i don't see the [environments passed in](https://github.com/monorist/monorise/blob/3b152178801c2ff5961949828ab5b717fdbbdc4b/packages/sst/components/monorise-core.ts#L174) has AWS_REGION, does that mean it will be all defaulted to `us-east-1`?

packages/sst/handlers/websocket.ts
```diff
@@ -0,0 +1,569 @@
...
+import type {
+  APIGatewayProxyEventV2,
+  APIGatewayProxyResultV2,
+  DynamoDBStreamEvent,
+  DynamoDBStreamHandler,
+} from 'aws-lambda';
+import { nanoid } from 'nanoid';
+
+const dynamodbClient = new DynamoDBClient({});
+const docClient = DynamoDBDocumentClient.from(dynamodbClient);
```
jyungtong: should import from dependencycontainer

packages/cli/cli.ts
wengkhing: Removed in 1063aaa.

packages/sst/index.ts
wengkhing: Cleaned up in 1063aaa.

packages/sst/components/monorise-core.ts
wengkhing: Good catch. Now uses `args.webSocket.handler?.memory` and `args.webSocket.handler?.timeout` with higher defaults (1024 MB / 60s) since broadcast processes DynamoDB streams. Fixed in 1063aaa.

packages/sst/handlers/websocket.ts
wengkhing: `AWS_REGION` is set automatically by Lambda runtime — the fallback is just a safety net. This entire function is replaced in #215 — `getWsEndpoint()` is removed and replaced with `WEBSOCKET_MANAGEMENT_ENDPOINT` env var passed directly from SST.

packages/sst/handlers/websocket.ts
wengkhing: WebSocket handlers are standalone Lambda functions invoked by API Gateway WebSocket, not Hono routes — they don't have access to DependencyContainer. Same pattern as the replication processor which also creates its own DynamoDB client.

jyungtong: TIL

wengkhing: @jyungtong Fixed the dependencycontainer issue, required more effort: https://github.com/monorist/monorise/pull/227

.changeset/happy-shrimps-kick.md
```diff
@@ -0,0 +1,34 @@
+---
+'@monorise/core': major
```
jyungtong: do we need major?

packages/sst/components/monorise-core.ts
```diff
@@ -149,6 +161,80 @@ export class MonoriseCore {
...
+     */
+    if (args?.webSocket?.enabled) {
+      const memory = args.webSocket.handler?.memory ?? '512 MB';
+      const timeout = args.webSocket.handler?.timeout ?? '30 seconds';
+
+      // WebSocket API Gateway
+      this.websocket = new sst.aws.ApiGatewayWebSocket(`${id}-websocket`, {});
+
+      const wsEnvironment = {
+        CORE_TABLE: this.table.table.name,
```
jyungtong, line 175: I understand we practise single table design, however, this websocket table doesn't seems relate to application data, wouldn't it looks like an infra data?

Should we split into Websocket dedicated table?

jyungtong: I also think from troubleshotting pov, we wanted to debug the subscription, but if we continue utilise app table, we can't have a glance straight because it's filled with app data, therefore we have to query with the exact key.

Also from performance pov, if websocket load is higher than app usage, we can tune websocket table dedicatedly.

jyungtong: if something gone wrong with websocket, we can simply delete every items in websocket table without worrying impacting app data. but if it sits in app data, we might not be that confident to run deletion quick, thus have to wait ttl

wengkhing: > I understand we practise single table design, however, this websocket table doesn't seems relate to application data, wouldn't it looks like an infra data?
> 
> Should we split into Websocket dedicated table?

@jyungtong It is related to app data, eg. if you are using monorise to store chat/conversation message, linear-like ticketing system. If we pull out to a different table, aren't it going to make "joining" table difficult? I would like to query all conversation for user A, all stories from epic B etc. It's up to the product whether real-time updates are required.

>  if something gone wrong with websocket, we can simply delete every items in websocket table without worrying impacting app data. but if it sits in app data, we might not be that confident to run deletion quick, thus have to wait ttl

I would treat it like replicator/mutual processor, if websocket is breaking and doesn't work, we shouldn't even be using/publishing it first.

.changeset/happy-shrimps-kick.md
wengkhing: Nice catch, we do not need major. I think my agentic workflow unknowingly treated some change within the same websocket epic as breaking changes. Will change to `minor` instead.

PR #203, opened 2026-03-27
feat(cli): one-command project scaffolding with npx monorise init

wengkhing:
## Summary

Documentation improvements and enhanced CLI init command for faster time-to-production.

### 1. Documentation Updates

Updated README.md and www/docs/ to reflect the new single-command init flow:
- Simplified quickstart to `npx monorise init --name my-app`
- Updated messaging to emphasize time-to-production
- Added "Ship in Hours" benefit
- Renamed `Prejoin` to `Tree` in README

### 2. Bug Fix

Fixed invalid `"version": null` in www/package.json causing Turborepo failures.

### 3. Enhanced CLI Init Command

`npx monorise init --name my-app` now creates a complete monorepo that supports multiple frontend apps and backend services:

- **Project scaffolding**: monorepo with `apps/`, `services/`, `monorise/configs/`
- **Next.js app** in `apps/web/` with Tailwind CSS
- **SST v4** infrastructure with `sst.config.ts`
- **Example User entity** with working frontend page (`useEntities`, `createEntity`)
- **Backend custom routes** with Hono template in `services/core/routes.ts`
- **Shadcn UI components**: button, card, input, label with theme variables
- **Global loader**: interruptive loading overlay via `useInterruptiveLoadStore`
- **Global initializer**: Monorise store configuration on app mount
- **API proxy routes**: `app/api/[...proxy]/route.ts` proxying to monorise backend
- **Path aliases**: `#/monorise`, `#/monorise/*`, and `#/*` configured in tsconfig
- **`init` is green-field only**: always scaffolds into a fresh directory named via `--name`; `--config-root` (used by `dev`/`build` to point at an existing project) does not apply here

### 4. Template Extraction

Extracted all inline template literals from `cli.ts` into `packages/cli/templates/` for:
- Better linting (template literals hide errors)
- Reduced `cli.ts` file size
- Easier template maintenance

### 5. Simplified import: `#/monorise`

`monorise dev`/`monorise build` now also generate `.monorise/index.ts` (re-exporting `.monorise/config.ts`), and the tsconfig gets an exact-match `'#/monorise'` path alias pointing at it. Generated (and existing) projects can now do:

```ts
import { Entity } from '#/monorise';
```

instead of `'#/monorise/config'` — the longer path keeps working unchanged (`config.ts` is still generated every build), so this is purely additive/backward compatible.

### 6. `createMutualConfig` scaffold sample

The scaffold now creates a second starter entity (`Team`) alongside `User`, plus `monorise/mutuals/team-membership.ts` — a shared `createMutualConfig` (`role: 'member' | 'admin'`) referenced from both entity configs, demonstrating a validated `User <-> Team` mutual relationship end to end.

### Screenshots

image

https://github.com/user-attachments/assets/b83d67ee-bde7-409d-9dbe-82842570d93d

## Review fixes

Addressed in follow-up commits after review:
- `updateEntity`/`init` no longer advertise `--config-root` for `init` (usage string + code both now green-field only, per above)
- Final "next steps" output now prints the actual detected page path instead of a hardcoded `apps/web/src/app/page.tsx` (scaffold uses `--no-src-dir`)
- `proxy-request.ts` no longer hardcodes `'secret1'` as `x-api-key` — reads `process.env.API_KEY`, sourced from `sst.config.ts`'s environment block (single place, matches the backend module's default `API_KEYS` secret)
- Removed a dead/broken `(:d+)?` regex group in the proxy URL rewrite
- Relabeled "SST v3" → "SST v4" everywhere (README, getting-started.md, changeset, cli.ts) to match the actual installed version
- Fixed `src/` path drift in README/getting-started.md (`--no-src-dir` means `apps/web/app/*`, not `apps/web/src/app/*`)
- Added `monorise` (the combined package) to the changeset — it has no npm `dependencies` entry on `@monorise/cli` (built via a file-copy script instead), so changesets' automatic dependency-cascade bump wouldn't have caught it

## Verification

- [x] Build passes
- [x] Tests pass
- [x] Tested init command creates working project
- [x] Verified `#/monorise` import + `createMutualConfig` scaffold end-to-end in a scratch project: codegen, `tsc --noEmit`, and `tsx` runtime resolution all succeed


www/docs/index.md
```diff
@@ -111,12 +111,12 @@ That's it — API, DynamoDB, EventBridge, processors, and a type-safe frontend.
...
-    <h3><span class="feature-icon">&#x1F512;</span> Type-Safe Schemas</h3>
-    <p>Define entities with Zod schemas. Get full TypeScript types across backend and frontend with zero code generation delay.</p>
+    <h3><span class="feature-icon">&#x1F504;</span> Zero Schema Drift</h3>
+    <p>One Zod config drives your entire stack. Edit &rarr; save &rarr; types regenerate instantly. DB, backend, and frontend stay perfectly in sync.</p>
   </div>
   <div class="feature">
-    <h3><span class="feature-icon">&#x26A1;</span> Event-Driven Processors</h3>
-    <p>Mutual, tag, and prejoin processors keep denormalized access patterns in sync via EventBridge and SQS.</p>
+    <h3><span class="feature-icon">&#x1F680;</span> Ship in Hours, Not Weeks</h3>
+    <p>From <code>git init</code> to production API in minutes. No migration scripts. No access pattern spreadsheets. No 2am GSI regrets.</p>
```
jyungtong: what meme is that `No 2am GSI regrets`

README.md
```diff
@@ -1,247 +1,165 @@
...
+                         ▼                              ▼
+                  ┌─────────────┐              ┌────────────────┐
+                  │  DynamoDB   │◀────────────▶│  EventBridge   │
+                  │ Single Table│              │     Bus        │
+                  └─────────────┘              └───────┬────────┘
+                                                       │
+                              ┌────────────────────────┼────────────────────────┐
+                              ▼                        ▼                        ▼
+                       ┌─────────────┐          ┌─────────────┐          ┌─────────────┐
+                       │ SQS Mutual  │          │  SQS Tag    │          │ SQS Prejoin │
```
jyungtong: `Prejoin` -> `Tree` ?

packages/cli/cli.ts
```diff
@@ -327,138 +336,427 @@ const MONORISE_LOGO = `
...
 });
 
 export default config;
 `;
-  if (!fs.existsSync(userEntityTsPath)) {
-    fs.writeFileSync(userEntityTsPath, userEntityContent.trimStart());
-    console.log(`Created ${path.relative(projectRoot, userEntityTsPath)}`);
+
+// Template for example Next.js page demonstrating monorise/react hooks
+const EXAMPLE_PAGE_TEMPLATE = `'use client';
```
jyungtong: shall we actually put these templates into a separate files, eg templates/app.ts .

The reason is because by having this inlined template, we can't tell if there's linting error because of it's template literal. Also making cli.ts bloater.

packages/cli/cli.ts
```diff
@@ -327,138 +336,427 @@ const MONORISE_LOGO = `
...
+  };
+  fs.writeFileSync(
+    path.join(projectRoot, 'services', 'core', 'package.json'),
+    JSON.stringify(servicesPackageJson, null, 2),
+  );
+
+  // Step 12: Create sst.config.ts with monorise
+  console.log('⚙️  Creating sst.config.ts with Monorise...');
+  const sstConfigPath = path.join(projectRoot, 'sst.config.ts');
+  const sstConfigContent = `/// <reference path='./.sst/platform/config.d.ts' />
```
jyungtong: same comment

www/docs/index.md
wengkhing: Reworded to "No hand-rolled GSI headaches." in 18dc0e6

README.md
wengkhing: Renamed to Tree in 18dc0e6

packages/cli/cli.ts
wengkhing: Done — extracted all templates into `packages/cli/templates/` in 18dc0e6. Also added new templates for shadcn UI components, global loader/initializer, API proxy routes, and globals.css.

packages/cli/cli.ts
wengkhing: Addressed together with the above.

packages/cli/templates/core-routes.ts
```diff
@@ -0,0 +1,23 @@
...
+  // Example: List users using the entity repository
+  const entities = await container.entityRepository.listEntities({
+    entityType: Entity.USER,
+  });
+
+  return c.json({ items: entities.items }, 200);
+});
+
+export default app;
+`;
```
jyungtong:
```suggestion
import { Hono } from 'hono';
import { DependencyContainer } from 'monorise/core';
import config, { Entity } from '#/monorise/config';

const container = new DependencyContainer(config);

const app = new Hono();

app.get('/health', async (c) => {
  return c.json({ status: 'ok', timestamp: new Date().toISOString() });
});

app.get('/test', async (c) => {
  // Example: List users using the entity repository
  const entities = await container.entityRepository.listEntities({
    entityType: Entity.USER,
  });

  return c.json({ items: entities.items }, 200);
});

export default app;
```

let's not use template literal, so the linting still works.
we use file reader to read this file and transform/copy to exact file during init.

eg snippet of init.ts
```
import fs from 'node:fs';
import path from 'node:path';
import { fileURLToPath } from 'node:url';

// Get the directory of the current file
const __dirname = path.dirname(fileURLToPath(import.meta.url));

export function initProject() {
  // 1. Resolve the path to your template file
  const templatePath = path.join(__dirname, '../templates/template.ts');

  // 2. Read it as a UTF-8 string (This does NOT execute the code)
  let templateContent = fs.readFileSync(templatePath, 'utf8');

  // 3. (Optional) Perform "Search and Replace" for dynamic templates
  templateContent = templateContent.replace('{{APP_NAME}}', 'MyCoolApp');

  // 4. Write it to the user's current working directory
  const outputPath = path.join(process.cwd(), 'app.ts');
  fs.writeFileSync(outputPath, templateContent);

  console.log('Successfully generated app.ts!');
}
```

wengkhing: Addressed in latest commit. All templates converted from exported string literals to plain source files in templates/files/. They're now real .ts/.tsx/.css files that get linted normally. At build time they're copied to dist/templates/files/ and read via fs.readFileSync at runtime. Dynamic values use {{PLACEHOLDER}} replaced with .replace().

Also updated the example page to use react-hook-form with zodResolver and FormSchema from the generated config instead of manual useState/isCreating state.

packages/cli/cli.ts
```diff
@@ -336,137 +355,324 @@ const MONORISE_LOGO = `
...
+  console.log(`\n📁 Project structure:`);
+  console.log(`  ${projectName}/`);
+  console.log(`  ├── apps/web/           # Next.js frontend`);
+  console.log(`  ├── services/core/      # Backend routes (Hono)`);
+  console.log(`  ├── monorise/configs/   # Entity definitions`);
+  console.log(`  └── .monorise/          # Generated files`);
+  console.log(`\n📂 Where to start coding:`);
+  console.log(`  • Edit your data model → monorise/configs/user.ts`);
+  console.log(`    (Add fields like phone, role, status to the User entity)`);
+  console.log(`  • Build your UI → apps/web/src/app/page.tsx`);
```
yeuharn96: [Claude] 🟡 bug: prints `apps/web/src/app/page.tsx`, but init runs `create-next-app --no-src-dir` (L418), so the page is actually written to `apps/web/app/page.tsx`. This final instruction points users at a path that doesn't exist.

packages/cli/cli.ts
```diff
@@ -336,137 +355,324 @@ const MONORISE_LOGO = `
...
     );
+  } catch (error) {
+    console.error('Failed to create Next.js app:', error);
+    process.exit(1);
+  }
+
+  // Step 5: Install SST v3
+  console.log('\n☁️  Installing SST v4...');
+  try {
+    execCommand('npm install sst@^4 --save-dev', projectRoot);
```
yeuharn96, line 441: [Claude] 🔵 nit: installs `sst@^4` (and the log at L427 says v4), but README L24/L129/L181, `sst.config.ts` comment, getting-started L49, and the changeset all say "SST v3". Config uses the ion API so v4 runs fine — just pick one label.

packages/cli/cli.ts
```diff
@@ -336,137 +355,324 @@ const MONORISE_LOGO = `
...
-export default config;
-`;
-  if (!fs.existsSync(userEntityTsPath)) {
-    fs.writeFileSync(userEntityTsPath, userEntityContent.trimStart());
-    console.log(`Created ${path.relative(projectRoot, userEntityTsPath)}`);
-  } else {
-    console.log(
-      `${path.relative(projectRoot, userEntityTsPath)} already exists. Skipping.`,
+  // Check if directory exists
+  if (fs.existsSync(projectRoot)) {
```
yeuharn96, line 391: [Claude] ❓ q: `init` now hard-exits when `projectRoot` exists, so `monorise init --config-root .` (or any existing dir) always fails — the old in-place scaffold is gone. Yet the usage string at L763 still advertises `--config-root` for `init`. If greenfield-only is intended, drop `--config-root` from the init usage line.

packages/cli/templates/files/proxy-request.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  const resp = await fetch(
+    rewriteUrl({ requestUrl: req.url, replacePath: path }),
+    {
+      method: req.method,
+      body,
+      headers: {
+        ...reqHeaders,
+        'x-api-key': 'secret1',
```
yeuharn96: [Claude] 🟡 risk: hardcoded `'x-api-key': 'secret1'` is baked into every scaffolded app and committed to the user's repo — every generated project ships the same key. Read from `process.env` (e.g. `process.env.API_KEY`) with no literal fallback.

packages/cli/templates/files/proxy-request.ts
```diff
@@ -0,0 +1,49 @@
...
+function rewriteUrl({
+  requestUrl,
+  replacePath,
+}: {
+  requestUrl: string;
+  replacePath?: string;
+}) {
+  const API_BASE_URL = process.env.API_BASE_URL;
+  const tobeReplacedPath =
+    replacePath ?? requestUrl.replace(/^https?:\/\/[^\/]+(:d+)?\/api\//, '');
```
yeuharn96: [Claude] 🔵 nit: `(:d+)?` matches a literal `d`, not a port — meant `(:\d+)?`. It's dead anyway since `[^/]+` already consumes the `:port`; drop it or fix the escape.

README.md
```diff
@@ -1,125 +1,186 @@
...
   baseSchema,
+  searchableFields: ['name', 'email'],
+  uniqueFields: ['email'],
 });
+```
+
+**3. Query from React:**
+
+```tsx
+// apps/web/src/app/page.tsx
```
yeuharn96: [Claude] 🔵 nit: `src/` paths throughout the README (here and the project tree ~L235) don't match the scaffold — `--no-src-dir` produces `apps/web/app/...`, not `apps/web/src/app/...`.

www/docs/getting-started.md
```diff
@@ -83,19 +104,54 @@ const config = createEntityConfig({
 export default config;
 ```
 
-Generate monorise artifacts from your config:
+### Frontend page (`apps/web/src/app/page.tsx`)
```
yeuharn96: [Claude] 🔵 nit: same `src/` drift — the tree (L38) and this heading show `apps/web/src/app/`, but `--no-src-dir` scaffolds `apps/web/app/`.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude**

Solid scaffolding upgrade — templates-as-files is the right call and the customRoutes/`appHandler` wiring checks out (raw Hono export is accepted at `handles/app.ts:31`). Requesting changes on a few drift/correctness items, mostly the `src/` path and SST version mismatches this branch is explicitly meant to clean up. Nothing blocks the happy path.

Inline comments below.

packages/cli/cli.ts
wengkhing: Fixed — final instructions now print the actual detected page path (via the same `pagePath` computed above for writing the example page) instead of a hardcoded `apps/web/src/app/page.tsx`, which never existed since scaffolding passes `--no-src-dir`.

packages/cli/cli.ts
wengkhing: Fixed — relabeled all "SST v3" mentions (README x4, getting-started.md, the changeset) to "SST v4", and fixed cli.ts's own internal inconsistency (the step comment said v3, the log said v4). Matches the actual installed version and `@monorise/sst`'s own peer dep (`^4.7.3`).

packages/cli/cli.ts
wengkhing: Confirmed green-field only — stopped wiring --config-root into runInitCommand entirely (it now always creates a fresh directory named via --name) and split the usage string per-command so it no longer advertises --config-root for init.

packages/cli/templates/files/proxy-request.ts
wengkhing: Fixed — proxy-request.ts now reads `process.env.API_KEY` with no hardcoded literal. Since `monorise.module.Core` doesn't currently expose its `API_KEYS` secret as a public property (checked `packages/sst/components/monorise-core.ts` — it's a local const in the constructor), sst.config.ts can't source it from the module output yet. Wired `API_KEY: 'secret1'` into the Next.js environment block instead — same value as the module's default `API_KEYS` secret, so a fresh scaffold still works out of the box, but now the literal lives in exactly one place (the project's own infra config, same pattern as API_BASE_URL) instead of being duplicated inside application code.

packages/cli/templates/files/proxy-request.ts
wengkhing: Fixed — dropped the dead group entirely. `[^/]+` already consumes `:port`, so it wasn't doing anything even before the typo.

README.md
wengkhing: Fixed — updated both instances (inline example + project tree) to match the --no-src-dir scaffold.

www/docs/getting-started.md
wengkhing: Fixed — updated the tree and the heading to apps/web/app/*.

wengkhing:
Pushed 9bd2952 addressing all review findings, plus confirming green-field-only per your answer:

- **`--config-root` for `init`**: confirmed green-field only. Stopped wiring `--config-root` into `runInitCommand` (it always creates a fresh dir named via `--name` now) and split the usage string per-command so it no longer advertises `--config-root` for `init`.
- **Wrong final-instructions path**: now prints the actual detected page path instead of a hardcoded `apps/web/src/app/page.tsx`.
- **Hardcoded API key**: `proxy-request.ts` now reads `process.env.API_KEY`, no literal fallback. Since `monorise.module.Core` doesn't expose its `API_KEYS` secret as a public property yet, wired `API_KEY: 'secret1'` into `sst.config.ts`'s environment block instead (matching the module's default secret) — so scaffolds still work out of the box, but the literal now lives in exactly one place instead of duplicated in application code.
- **Dead/broken regex**: dropped `(:d+)?` — `[^/]+` already consumed the port.
- **SST v3/v4 labeling**: relabeled everywhere to v4, matching the actual installed version and `@monorise/sst`'s peer dep.
- **`src/` path drift**: fixed in README and getting-started.md — `--no-src-dir` means `apps/web/app/*`, not `apps/web/src/app/*`.

Build passes (pre-existing, unrelated warnings only).

yeuharn96 (approved):

PR #204, opened 2026-03-27
Fix combined package DTS rewriting and CLI monorepo detection

jovanimal:
### Bug 1: DTS import rewriting misses @monorise/* references (build.js)

  The regex in build.js failed to rewrite all @monorise/* imports to relative paths in.d.ts files due to:
  1. .test() advancing global regex lastIndex before .replace(), skipping the first match
  2. Only matching single-quoted patterns — missing import("@monorise/base") (double
  quotes)

  Fix: Direct .replace() calls for both quote styles, simple string comparison for change
  detection.

### Bug 2: CLI fails to detect combined package in monorepos (cli.ts)

  The CLI checked projectRoot/node_modules/monorise to detect the combined package, but in workspaces dependencies are hoisted to root node_modules/. This caused wrong import paths in handle.ts and TS2664 errors from invalid module augmentation blocks.

  Fix: Added detectCombinedPackage() that walks up the directory tree (mirroring Node.js resolution). The CLI now generates only the relevant module augmentation based on detection — monorise/base for combined, @monorise/base for scoped.
  
  
image

### DIFF
--- a/.changeset/fix-combined-package-dts-and-cli.md
+++ b/.changeset/fix-combined-package-dts-and-cli.md
@@ -1,0 +1,9 @@
+---
+"@monorise/cli": patch
+"monorise": patch
+---
+
+Fix combined package DTS rewriting and CLI monorepo detection
+
+- build.js: Fix regex that missed rewriting some `@monorise/*` imports in `.d.ts` files (global regex `lastIndex` bug + missing double-quote patterns)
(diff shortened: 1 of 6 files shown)


jovanimal:
am merging it for now because it's breaking the build when i use the combined package `0.0.4`, let me know if it's not the case for you.

PR #214, opened 2026-04-05
Support array fields in searchableFields

jovanimal:
## Summary
- Handle array values in `searchableFields` by flattening elements into the search string
- Normalize dashes and underscores to spaces for all searchable fields, enabling searches like "shah alam" to match "shah-alam"

### DIFF
--- a/packages/core/data/Entity.ts
+++ b/packages/core/data/Entity.ts
@@ -767,9 +767,15 @@ export class EntityRepository extends Repository {
 
     for (const item of results.items) {
       const searchTerm = (searchableFields ?? [])
-        .map((field) =>
-          (item.data as Record<string, any>)[field]?.toLowerCase(),
-        )
+        .map((field) => {
+          const value = (item.data as Record<string, unknown>)[field];
+          if (Array.isArray(value)) {
+            return value
+              .map((v) => String(v).replace(/[-_]/g, ' ').toLowerCase())
+              .join(' ');
+          }
+          return String(value ?? '').replace(/[-_]/g, ' ').toLowerCase();
+        })
         .join(' ');
       const isMatched = queryRegex.test(searchTerm);
       if (isMatched) {
--- a/.changeset/search-array-fields.md
+++ b/.changeset/search-array-fields.md
@@ -1,0 +1,5 @@
+---
+"@monorise/core": patch
+---
+
+Support array fields in searchableFields and normalize dashes/underscores to spaces for better search matching


wengkhing:
@jovanimal Something I keen to add to `searchableFields`, see if you are interested to solve this or not.

Currently, searchableFields are probably the worst performing feature in monorise. What happen when we try to search for a term:
1. list all entities
2. find if any entity records that matches the search term

This is not great because it requires scanning the entire entity records, say if you've 100k records, this is going to pump up read cost a lot.

My proposal:
split strings in the field into words and index them by word:

Example:
PK = SEARCH#user#fullname, SK = {{broken_down_search_term}}, GS1PK = user#{{id}}
This is a row of replicated data based on PK = user#{{id}}, SK = #METADATA#

Meaning, we could get the exact record based on search term instead of scanning through the entire entity records.

cc @jyungtong

jovanimal:
@wengkhing interesting. sure, let me see what's the most optimal approach to tackle this

packages/core/data/Entity.ts
```diff
@@ -767,9 +767,15 @@ export class EntityRepository extends Repository {
...
-          (item.data as Record<string, any>)[field]?.toLowerCase(),
-        )
+        .map((field) => {
+          const value = (item.data as Record<string, unknown>)[field];
+          if (Array.isArray(value)) {
+            return value
+              .map((v) => String(v).replace(/[-_]/g, ' ').toLowerCase())
+              .join(' ');
+          }
+          return String(value ?? '').replace(/[-_]/g, ' ').toLowerCase();
```
jyungtong, line 777: I think the regex replacement is more like an application layer requirements than an infra layer requirement.

I suggest we can normalize search term in the app input before passing to query params instead.

packages/core/data/Entity.ts
```diff
@@ -767,9 +767,15 @@ export class EntityRepository extends Repository {
 
     for (const item of results.items) {
       const searchTerm = (searchableFields ?? [])
-        .map((field) =>
-          (item.data as Record<string, any>)[field]?.toLowerCase(),
-        )
+        .map((field) => {
+          const value = (item.data as Record<string, unknown>)[field];
+          if (Array.isArray(value)) {
```
jyungtong, line 772: i agree to have this requirement, probably just not the regex

packages/core/data/Entity.ts
jovanimal: @jyungtong last week i was thinking the same thing to keep the logic in the app, before I come to this proposal. I actually asked Opus: 

> i am thinking whether having another field in our `createEntityConfig` method will be a better approach here as that will allow us to have more control. For example, we can introduce a new field on top of what we already have `searchableFields`, maybe call `searchableFieldsRegex` and  then we can define our own regex in there, in order to make  the search function more powerful and have a better match.

Answer:

Image

Come to think of this, this normalization here is rather universal and shouldn't tie to the app logic, most of the time we will use `-` or `_` in our codebase for longer strings, if HumanThis today has `actorAttributes` = `['long-hair', 'double-chin']`, then users are able to search in a more intuitive way, e.g. `long hair`

@wengkhing what do you think about this?

jovanimal:
@wengkhing on your proposal earlier on improving search mechanism, 

Potential concerns:
  - Write amplification: A record with 10 searchable fields averaging 5 words each = ~50 index rows per entity. Every create/update/delete must maintain all those rows
   - Partial matching is limited: begins_with on SK only matches from the start of a word — searching "lumbing" won't match "plumbing"
  - Multi-word queries are complex: Searching "bp plumbing" requires querying "bp" and "plumbing" separately, then intersecting results in application code
  - Storage cost: Every word in every searchable field gets its own DynamoDB row
  
The suggestions are mostly to implement a third-party search engine like OpenSearch, Algolia etc. I have also tried to research whether there is any method that we can just amend DynamoDB's query method today without implementing 3rd-party search engine, apparently there isn't any clear win without tradeoffs.

wengkhing:
@jovanimal Yeah, these came to my mind before the proposal, here's my thought:

> * Write amplification: A record with 10 searchable fields averaging 5 words each = ~50 index rows per entity. Every create/update/delete must maintain all those rows

Agree, it would be more costly to write, depends on how many words are there.

> * Partial matching is limited: begins_with on SK only matches from the start of a word — searching "lumbing" won't match "plumbing"

I have this concern as well, it seems very unlikely when we wanna search for `plumbing`, but we input `lumbing`, just personal gut feeling. Therefore I also have an idea of using tokenizer like `natural`: https://naturalnode.github.io/natural/Tokenizers.html

But I feel like might be overkill and increasing even more write per record. However, I think language like Malay make more sense:

> * Multi-word queries are complex: Searching "bp plumbing" requires querying "bp" and "plumbing" separately, then intersecting results in application code

Yup, exactly what I thought, intersecting result when space involved.

> * Storage cost: Every word in every searchable field gets its own DynamoDB row

Storage cost is probably the cheapest cost and least concern among these all points.

> The suggestions are mostly to implement a third-party search engine like OpenSearch, Algolia etc. I have also tried to research whether there is any method that we can just amend DynamoDB's query method today without implementing 3rd-party search engine, apparently there isn't any clear win without tradeoffs.

Totally not against the idea of using better tools like OpenSearch. OpenSearch requires a 24/7 running instance as opposed to DynamoDB can be charged on-demand. If you anticipate there's a lot of search traffic, OpenSearch might be a better option. https://aws.amazon.com/opensearch-service/pricing/

I think this whole idea should toggle-able, depends on whether if your app has use case for this.

PR #217, opened 2026-04-09
Websocket #4 - feat(react): add useEntityFeed hook

wengkhing:
## Summary

Phase 3 of the entity feed feature — the React hook that manages ticket lifecycle, WebSocket connection, and store updates.

## Changes

- `useEntityFeed({ entityType, entityId, ticketEndpoint? })` hook
- Default ticket endpoint: `POST /api/core/ws/ticket/:entityType/:entityId` (catch-all proxy)
- Custom endpoint: developer's own proxy route with their auth
- Routes `entity.created/updated/deleted` and `mutual.created/updated/deleted` into zustand stores
- Proactive ticket refresh (2 min before expiry) with reactive fallback
- Sets `globalWsManager` so `useMutualSocket`, `useEphemeralSocket` etc. coexist
- `UseEntityFeedOptions` and `UseEntityFeedReturn` exported interfaces
- `httpActions` parameter made optional in `initWebSocketActions`

### Reconnect handling

- **Stable disconnect** (connected >5s then dropped): auto-reconnect with exponential backoff (2s→4s→8s→16s→32s), max 5 attempts
- **Unstable disconnect** (connected <5s): no reconnect — prevents infinite loop
- **Network recovery**: browser `online` event triggers immediate reconnect (skips if already connected/connecting)
- **Sleep/wake**: periodic interval (5s) detects >30s elapsed time, force-reconnects with fresh ticket
- **Stale manager**: identity check (`wsManagerRef.current !== manager`) ignores events from old managers
- **Concurrent guard**: `connectingRef` prevents double-connect from rapid events or React strict mode
- **Ticket retry**: 3 attempts with 1s→2s→4s backoff on transient errors, auth errors fail immediately

## Depends on

- #216

## Test plan

- [ ] Verify `useEntityFeed` fetches ticket and connects via WebSocket
- [ ] Verify entity store updates when entity broadcast arrives
- [ ] Verify mutual store updates when mutual broadcast arrives
- [ ] Verify ticket refreshes before expiry
- [ ] Turn off wifi → verify reconnect attempts with backoff → turn on wifi → verify recovers to "Live"
- [ ] Sleep laptop >30s → wake → verify reconnects
- [ ] Verify `TICKET_UNAUTHORIZED` error surfaces when proxy returns 401
- [ ] Verify no reconnect loop when connection drops immediately

### DIFF
--- a/packages/react/actions/websocket.action.ts
+++ b/packages/react/actions/websocket.action.ts
@@ -494,6 +494,8 @@ export const initWebSocketActions = (
     const wsManagerRef = useRef<WebSocketManager | null>(null);
     const isMountedRef = useRef(true);
     const connectingRef = useRef(false);
+    const reconnectCountRef = useRef(0);
+    const lastConnectedAtRef = useRef(0);
 
     const fetchTicket = useCallback(async () => {
       try {
@@ -585,11 +587,26 @@ export const initWebSocketActions = (
 
         if (state === 'connected') {
           connectingRef.current = false;
+          reconnectCountRef.current = 0;
+          lastConnectedAtRef.current = Date.now();
           setIsConnected(true);
           setError(null);
         } else if (state === 'disconnected') {
           connectingRef.current = false;
           setIsConnected(false);
+
+          // Only auto-reconnect if we were connected for >5s (stable connection)
+          // and haven't exceeded max reconnect attempts
+          const wasStable = Date.now() - lastConnectedAtRef.current > 5000;
+          if (wasStable && reconnectCountRef.current < 5) {
+            reconnectCountRef.current++;
+            const delay = 2000 * Math.pow(2, reconnectCountRef.current - 1);
+            setTimeout(() => {
+              if (isMountedRef.current && !connectingRef.current) {
+                connectRef.current();
+              }
+            }, delay);
+          }
         }
       });
 
@@ -681,25 +698,35 @@ export const initWebSocketActions = (
       }
 
       // Detect wake from sleep using periodic check
-      // visibilitychange doesn't reliably fire on laptop sleep
       let lastCheckTime = Date.now();
       const sleepCheckInterval = setInterval(() => {
         const now = Date.now();
         const elapsed = now - lastCheckTime;
         lastCheckTime = now;
 
-        // If >30s elapsed since last check, device was likely asleep
         if (elapsed > 30000 && wsManagerRef.current) {
           wsManagerRef.current.disconnect();
           wsManagerRef.current = null;
           connectingRef.current = false;
+          reconnectCountRef.current = 0;
           connectRef.current();
         }
       }, 5000);
 
+      // Reconnect when network comes back online
+      const handleOnline = () => {
+        if (!isMountedRef.current) return;
+        if (wsManagerRef.current?.getState() === 'connected') return;
+        if (connectingRef.current) return;
+        reconnectCountRef.current = 0;
+        connectRef.current();
+      };
+      window.addEventListener('online', handleOnline);
+
       return () => {
         isMountedRef.current = false;
         clearInterval(sleepCheckInterval);
+        window.removeEventListener('online', handleOnline);
         if (refreshTimerRef.current) {
           clearTimeout(refreshTimerRef.current);
         }
--- a/packages/react/actions/websocket.action.ts
+++ b/packages/react/actions/websocket.action.ts
@@ -95,6 +95,7 @@ export const initWebSocketActions = (
     });
 
     const fetchData = async (type: 'initial' | 'more' | 'refresh') => {
+      if (!httpActions) return;
       if (type === 'more') {
         setIsFetchingMore(true);
       } else if (type === 'refresh') {
@@ -106,7 +107,7 @@ export const initWebSocketActions = (
 
       try {
         const fetchLastKey = type === 'refresh' ? undefined : lastKeyRef.current;
-        const result = await httpActions!.listEntities(entityType, {
+        const result = await httpActions.listEntities(entityType, {
           limit,
           lastKey: fetchLastKey,
         });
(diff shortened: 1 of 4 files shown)


packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+        }
+
+        const data = await response.json();
+        return { data: data as { ticket: string; wsUrl: string; expiresIn: number } };
+      } catch (err) {
+        return { error: { code: 'TICKET_FAILED', message: (err as Error).message } };
+      }
+    }, [entityType, entityId, ticketEndpoint]);
+
+    const connectWithTicket = useCallback(async () => {
```
shubhamkakkar, line 534: `useEntityFeed` creates a `connectWithTicket` function with a complex dependency array in every render, and it's also wrapped in a `useCallback` that references `fetchTicket`. This might be overkill if `connectWithTicket` is assigned to a `ref`. Consider moving the `WebSocketManager` instantiation and logic into a more stable structure or custom hook to reduce the dependency footprint.

packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+          return { error: { code: 'TICKET_UNAUTHORIZED', message: 'Unauthorized' } };
+        }
+
+        if (!response.ok) {
+          return { error: { code: 'TICKET_FAILED', message: `HTTP ${response.status}` } };
+        }
+
+        const data = await response.json();
+        return { data: data as { ticket: string; wsUrl: string; expiresIn: number } };
+      } catch (err) {
```
shubhamkakkar, line 529: The catch block swallows the error and returns it as a stringified code/message pair. It should be logged via the project's logger to ensure observability for debugging failed production requests.

packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+   * // Internal app (uses catch-all proxy)
+   * useEntityFeed({ entityType: Entity.USER, entityId: userId });
+   *
+   * // Client-facing app (custom auth endpoint)
+   * useEntityFeed({ entityType: Entity.USER, entityId: userId, ticketEndpoint: '/api/ws/ticket' });
+   * ```
+   */
+  const useEntityFeed = (opts: UseEntityFeedOptions): UseEntityFeedReturn => {
+    const { entityType, entityId, ticketEndpoint } = opts;
+    const [isConnected, setIsConnected] = useState(false);
```
shubhamkakkar, line 492: `useEntityFeed` performs a `fetch` and multiple `setState` updates. Consider using a single `useReducer` to manage `isConnected`, `error`, and `reconnectCount` to ensure state transitions are batched and atomic, preventing multiple intermediate re-renders.

packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+    useEffect(() => {
+      isMountedRef.current = true;
+
+      if (entityId) {
+        connectRef.current();
+      }
+
+      // Detect wake from sleep using periodic check
+      let lastCheckTime = Date.now();
+      const sleepCheckInterval = setInterval(() => {
```
shubhamkakkar, line 703: The `sleepCheckInterval` uses `setInterval` to perform cleanup and reconnection. This is prone to race conditions if the connection process takes longer than the interval or if multiple intervals overlap. Consider using a `setTimeout`-based recursive polling loop or checking the status of the `wsManagerRef` before blindly disconnecting.

packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+
+        if (response.status === 401 || response.status === 403) {
+          return { error: { code: 'TICKET_UNAUTHORIZED', message: 'Unauthorized' } };
+        }
+
+        if (!response.ok) {
+          return { error: { code: 'TICKET_FAILED', message: `HTTP ${response.status}` } };
+        }
+
+        const data = await response.json();
```
shubhamkakkar, line 527: The `try/catch` block for `response.json()` is missing. If the API returns a non-JSON body (or an empty one), `response.json()` will throw, which is currently unhandled here. Add a check for response content-type or wrap in `try/catch`.

packages/react/actions/websocket.action.ts
```diff
@@ -459,10 +471,281 @@ export const initWebSocketActions = (
...
+          connectingRef.current = false;
+          setIsConnected(false);
+
+          // Only auto-reconnect if we were connected for >5s (stable connection)
+          // and haven't exceeded max reconnect attempts
+          const wasStable = Date.now() - lastConnectedAtRef.current > 5000;
+          if (wasStable && reconnectCountRef.current < 5) {
+            reconnectCountRef.current++;
+            const delay = 2000 * Math.pow(2, reconnectCountRef.current - 1);
+            setTimeout(() => {
```
shubhamkakkar, line 606: There is no cleanup for the `setTimeout` initiated inside the `state === 'disconnected'` block. If the component unmounts while a reconnection delay is pending, it could still attempt to `connectRef.current()` if `isMountedRef` isn't checked correctly or if the timeout fires just as the component unmounts. Ensure the timeout is tracked in a `ref` and cleared in the `useEffect` cleanup.

packages/react/index.ts
```diff
@@ -6,7 +6,7 @@ import { initConfigActions } from './actions/config.action';
 import { initCoreActions } from './actions/core.action';
 import {
   initWebSocketActions,
-  initWebSocket,
+  initializeWebSocketManager,
```
shubhamkakkar, line 9: I see `initializeWebSocketManager` is being exported. In this file, all other initialization functions (like `initAppActions`) follow an `init...` naming convention. For consistency across the API surface, please rename this to `initWebSocketManager` to match the surrounding codebase.

PR #218, opened 2026-04-09
feat(websocket): connection cleanup via R1 GSI

wengkhing:
## Summary

Phase 4 of the entity feed feature — proper cleanup of subscription records on disconnect using the R1 GSI.

- Previously, `$disconnect` only deleted the connection record. Subscription records were left orphaned
- The initial approach tracked `subscriptionKeys[]` on the connection record — this was a workaround that required an extra `UpdateCommand` on every subscribe
- Now uses R1 GSI (`ENTITY_REPLICATION_INDEX`) for reverse lookup: all subscription records set `R1PK = CONN#{connectionId}`, so disconnect queries the GSI to find and delete them all

## Why R1 GSI?

R1 GSI already exists for entity data replication. Subscription records set `R1PK = CONN#{connectionId}` which doesn't collide with entity records (they use `R1PK = EntityType#EntityId`). The GSI projection includes PK and SK, which is all we need to construct delete keys.

This eliminates:
- The `subscriptionKeys[]` field on connection records
- The `trackSubscription()` helper and `UpdateCommand` import
- An extra DynamoDB write on every subscribe operation

## Changes

- `$disconnect` queries `ENTITY_REPLICATION_INDEX` with `R1PK = CONN#{connectionId}`
- Deletes all subscription records found + the connection record, in parallel
- Remove `trackSubscription()` function
- Remove `UpdateCommand` import

## Depends on

- #217

## Test plan

- [ ] Subscribe to entity, mutual, ephemeral, and feed types
- [ ] Verify R1PK/R1SK are set on all subscription records
- [ ] Disconnect and verify all subscription records are deleted
- [ ] Verify connection record is deleted
- [ ] Verify no orphaned subscription records remain

### DIFF
--- a/.changeset/entity-feed-hook.md
+++ b/.changeset/entity-feed-hook.md
@@ -1,0 +1,13 @@
+---
+"@monorise/react": minor
+---
+
+Add `useEntityFeed` hook for graph-aware real-time subscriptions.
+
+- Ticket-based WebSocket connection (default route or custom ticketEndpoint)
+- Auto-routes entity/mutual broadcast events into zustand stores
+- Proactive ticket refresh before expiry
+- Periodic sleep detection with automatic reconnect
+- Ticket fetch retry with exponential backoff on transient errors
+- Stale manager identity check prevents reconnect loops
+- React strict mode safe via connectingRef guard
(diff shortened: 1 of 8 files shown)


jyungtong:
no file changes?

wengkhing:
> no file changes?

@jyungtong Closing this PR, the changes are actually in another PR: https://github.com/monorist/monorise/pull/216/changes#diff-873ca74bfe0dfaf7a9a0d52a5ffff1cca4ee9ed7591a082ebf4f31899c3d0532

PR #228, opened 2026-04-14
chore: upgrade SST support to v4

wengkhing:
## Summary

Upgrades Monorise's SST integration from v3 to v4.

## Why

SST v4 ships with an upgraded Pulumi AWS provider (v6 → v7) and new features. Staying on v3 blocks users from adopting the latest SST ecosystem.

### Dependencies
- Bumped `sst` peer dependency from `^3.16.3` to `^4.0.0` in `@monorise/sst` and the unified `monorise` package
- Bumped internal `@monorise/base` references to `^4.0.0` in `core`, `react`, and `proxy`

### Code Fixes
- **Fixed type leak** in `packages/sst/components/q-function.ts` that imported `DurationMinutes` and `Input` from `../.sst/platform/src/components/*`. This path doesn't exist for consumers and breaks builds.
- **Fixed WebSocket handler** in `packages/sst/handlers/websocket.ts` to cast `event.requestContext.connectionId` for compatibility with updated `aws-lambda` types under SST v4.

### Examples
- `examples/basic`: switched from `@monorise/sst` to the unified `monorise` package; bumped `sst` to `^4.0.0`
- `examples/websocket-chat`: bumped `sst` to `^4.0.0` and `monorise` to the latest local build

### Documentation
- Added `docs/MIGRATE-SST-V4.md` with step-by-step migration instructions

### Release
- Added changeset for major version bumps (handled by Changesets release workflow)

## Migration Path

Users on Monorise v0 / SST v3 should:
1. `npm install sst@^4.0.0 monorise@^1.0.0` (or `@monorise/sst@^4.0.0`)
2. `npx sst install`
3. `npx sst diff` — verify no DynamoDB `replace`
4. `npx sst refresh` — one-way state migration
5. `npx sst deploy`

No application code changes are required.

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -2,15 +2,14 @@
   "name": "basic",
   "private": true,
   "type": "module",
+  "dependencies": {
+    "monorise": "file:../../packages/monorise/monorise-1.0.0.tgz",
+    "sst": "^4.7.3",
+    "zod": "^3.25.67"
+  },
   "devDependencies": {
     "@types/bun": "latest",
-    "@monorise/cli": "1.0.0-dev.1",
-    "@monorise/sst": "1.0.0-dev.2"
-  },
-  "peerDependencies": {
-    "typescript": "^5",
-    "@monorise/base": "1.0.0-dev.0",
-    "@monorise/core": "1.0.0-dev.0"
+    "typescript": "^5"
   },
   "scripts": {
     "dev": "sst dev"
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -3,8 +3,8 @@
   "private": true,
   "type": "module",
   "dependencies": {
-    "monorise": "file:../../packages/monorise/monorise-1.0.0.tgz",
-    "sst": "^4.7.3",
+    "monorise": "file:../../packages/monorise",
+    "sst": "4.7.3",
     "zod": "^3.25.67"
   },
   "devDependencies": {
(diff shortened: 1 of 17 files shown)


examples/basic/package.json
```diff
@@ -2,15 +2,14 @@
   "name": "basic",
   "private": true,
   "type": "module",
+  "dependencies": {
+    "monorise": "file:../../packages/monorise",
```
jyungtong, line 6: should we use version instead?

packages/sst/components/q-function.ts
```diff
@@ -43,7 +47,7 @@ export class QFunction {
     this.queue.subscribe(this.function.arn, {
       batch: {
         partialResponses: true,
-        window: maxBatchingWindow,
+        window: maxBatchingWindow as any,
```
jyungtong: why not `Input<DurationMinutes>`

packages/sst/handlers/websocket.ts
```diff
@@ -68,7 +73,7 @@ const getWsEndpoint = () => {
  * $connect handler
```
jyungtong, line 73: should use `APIGatewayProxyWebsocketEventV2`, so don't have to cast to any

examples/basic/package.json
wengkhing: This has to update again after publishing the new version merging this PR

packages/sst/handlers/websocket.ts
wengkhing: Yeah, was pushing my fix when you commenting this, already addressed

packages/sst/components/q-function.ts
wengkhing: Updated. No longer casting the type

PR #230, opened 2026-04-14
fix(changeset): remove @monorise/proxy from upgrade-sst-v4 changeset

wengkhing:
The release action is failing because `@monorise/proxy` does not exist in the main workspace yet (it was introduced as part of websocket features). This removes it from the changeset so the release can proceed.

### DIFF
--- a/.changeset/upgrade-sst-v4.md
+++ b/.changeset/upgrade-sst-v4.md
@@ -2,7 +2,6 @@
 "@monorise/base": major
 "@monorise/cli": major
 "@monorise/core": major
-"@monorise/proxy": major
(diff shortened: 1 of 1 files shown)


jyungtong (approved):

.changeset/upgrade-sst-v4.md
```diff
@@ -2,7 +2,6 @@
 "@monorise/base": major
 "@monorise/cli": major
 "@monorise/core": major
-"@monorise/proxy": major
```
jyungtong, line 5: Which model is this that hallucinated

PR #223, opened 2026-04-13
Add conditional where support for entity PATCH updates

jyungtong:
## Summary
- add optional top-level `where` support in `PATCH /core/entity/:entityType/:entityId` to enable atomic compare-and-set style updates
- compile service-layer `where` clauses into DynamoDB `ConditionExpression` options passed into repository updates
- map conditional failures to `CONDITIONAL_CHECK_FAILED` and return HTTP 409 in the update controller while preserving existing non-conditional behavior
- add comprehensive repository tests for supported operators and document conditional PATCH usage in README

## Testing
- npm -w @monorise/core run test

PR #231, opened 2026-04-14
feat(react,base): expose conditional PATCH updates via $where

jyungtong:
## Summary

- Add shared `WhereOperator`, `WhereClause`, `WhereConditions` types to `@monorise/base`
- Thread `where?: WhereConditions` param through `editEntity` in react's coreService and coreActions
- Merge conditions into PATCH body as `$where` field
- Re-export where types from `@monorise/react`

Exposes conditional update support (from #223) to React package consumers. Allows callers to guard entity updates with server-side conditions.

## Test plan

- [x] Both packages build successfully
- [x] Types exported from `@monorise/react`
- [x] `editEntity` accepts optional `where` parameter
- [x] Conditions merged into PATCH body as `$where`

Related: #223

### DIFF
--- a/.changeset/react-conditional-patch-where.md
+++ b/.changeset/react-conditional-patch-where.md
@@ -1,0 +1,10 @@
+---
+"@monorise/base": minor
+"@monorise/react": minor
+---
+
+Expose conditional PATCH updates via `$where` in react package
+
+- Added `WhereOperator`, `WhereClause`, `WhereConditions` types to `@monorise/base`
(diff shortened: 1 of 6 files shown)


.changeset/react-conditional-patch-where.md
wengkhing, line 1: Need to bump `monorise` version as well so unified package gets the update

jyungtong: 7df1214

wengkhing:
Work continued here: https://github.com/monorist/monorise/pull/234

PR #234, opened 2026-04-20
Transaction #1 - feat: unified named conditions for conditional writes

wengkhing:
## Summary

Introduces named conditions for entity write operations, replacing two mechanisms with limitations:

**Problem 1 — Security:** `updateEntity` exposed raw `$where` operators to frontend clients. An authenticated client could probe field existence or guess values via status codes. DynamoDB condition operators should never be client-facing.

**Problem 2 — Inflexibility:** `adjustEntity` used `adjustmentConstraints` with rigid static `min`/`max` or `minField`/`maxField` patterns. No way to express conditions based on external config (e.g., platform limits) or arbitrary field checks beyond numeric bounds.

**Solution:** Two server-defined named condition configs — `adjustmentConditions` for adjustEntity, `updateConditions` for updateEntity. Client sends a condition name (`$condition: 'withdraw'`), server resolves it to a DynamoDB ConditionExpression.

### Key changes

- **`adjustmentConditions`** config — static `WhereConditions` or `(data, adjustments) => WhereConditions` functions. `$condition` **required** when defined (unguarded deltas are dangerous)
- **`updateConditions`** config — static `WhereConditions` or `(data) => WhereConditions` functions. `$condition` always **optional**
- **Condition resolver** (`resolve-condition.ts`) — separate `resolveAdjustmentCondition` and `resolveUpdateCondition` functions
- **`INVALID_CONDITION` error** — returned as 400 for missing/unknown condition names
- **Backward compatibility**: legacy `$where` on updateEntity still works (with deprecation warning), legacy `adjustmentConstraints` still works when no `adjustmentConditions` defined
- **React SDK**: `adjustEntity` accepts `condition` in opts, `editEntity` accepts `$condition` in values

### Example

```typescript
createEntityConfig({
  name: 'wallet',
  baseSchema,
  adjustmentConditions: {
    withdraw: (data, adjustments) => ({
      balance: { $gte: (data.minBalance ?? 0) + Math.abs(adjustments.balance ?? 0) },
    }),
    deposit: { balance: { $lte: platformConfig.maxBalance } },
  },
  updateConditions: {
    publish: { status: { $eq: 'draft' } },
    archive: (data) => ({ status: { $ne: 'archived' } }),
  },
})

// Client sends condition name, never raw operators
POST /entity/wallet/:id/adjust   { balance: -10, $condition: 'withdraw' }
PATCH /entity/wallet/:id         { status: 'published', $condition: 'publish' }
```

## Test plan

- [x] All 101 existing + new tests pass (6 test files)
- [x] New HTTP-level tests for `$condition` on adjustEntity (8 tests): required condition, static/dynamic conditions, violation → 409, unknown name → 400, concurrent safety
- [x] New HTTP-level tests for `$condition` on updateEntity (5 tests): static match/mismatch, unknown name, no-conditions-defined backward compat
- [x] Existing `$where` tests still pass (backward compat)
- [x] Existing adjustEntity constraint tests converted to new API
- [x] Build succeeds across all packages
- [ ] Verify ledger example wallet config works with new conditions syntax


packages/core/services/entity.service.ts
```diff
@@ -186,10 +230,39 @@ export class EntityService {
...
+        }
+        opts = await resolveUpdateCondition({
+          conditionName: condition,
+          conditions: updateConditions,
+          getEntityData: async () => {
+            const entity = await this.entityRepository.getEntity(entityType, entityId);
+            return entity?.data ?? {};
+          },
+        });
+      } else if (where && Object.keys(where).length > 0) {
```
yeuharn96, line 268: 🟡 risk (security): raw `$where` is still accepted straight from the client body and passed to `buildConditionExpression` here, guarded only by a `console.warn`. The whole premise of this PR is that DynamoDB condition operators must never be client-facing — an authenticated client can still send `{ $where: { secretField: { $exists: true } } }` (or `$eq` value guesses) and read field existence / values off the 200-vs-409 status. Keeping `$where` for backward compat is reasonable, but it needs to be *opt-in on the server* (e.g. a config flag like `allowLegacyWhere`), defaulting off, otherwise the vulnerability the PR sets out to fix is still shipped. At minimum, document that leaving `$where` enabled re-exposes the original issue.

examples/ledger/monorise/configs/wallet.ts
```diff
@@ -18,8 +18,11 @@ const config = createEntityConfig({
   displayName: 'Wallet',
   baseSchema,
   createSchema,
-  adjustmentConstraints: {
-    balance: { min: 0 },
+  adjustmentConditions: {
+    withdraw: (data, adjustments) => ({
+      balance: { $gte: Math.abs(adjustments.balance ?? 0) },
+    }),
+    deposit: { balance: { $lte: 1000000 } },
```
yeuharn96: 🟡 risk: `deposit: { balance: { $lte: 1000000 } }` checks the balance *before* the delta is applied, so it doesn't cap the post-deposit balance. Depositing 500k onto a balance of 999,999 passes (`999999 <= 1000000`) and lands at 1,499,999 — the cap is exceeded. The old `adjustmentConstraints` avoided this by comparing against `max - delta`. Note `withdraw` right above already accounts for the delta via `abs(adjustments.balance)` — `deposit` should be a function too:
```ts
deposit: (data, adjustments) => ({
  balance: { $lte: 1000000 - (adjustments.balance ?? 0) },
}),
```
Same issue in the `deposit` fixture in `test-utils.ts` — worth a test that deposits *up to and past* the cap to lock the semantics in. This is the canonical example devs will copy, so the docs example in `www/docs/react.md` should match.

packages/core/services/entity.service.ts
```diff
@@ -104,17 +110,51 @@ export class EntityService {
...
+      opts = await resolveAdjustmentCondition({
+        conditionName: condition,
+        conditions: adjustmentConditions,
+        adjustments,
+        getEntityData: async () => {
+          const entity = await this.entityRepository.getEntity(entityType, entityId);
+          return entity?.data ?? {};
+        },
+      });
+    } else if (rawConstraints) {
```
yeuharn96, line 160: 🟡 risk: when `adjustmentConditions` is undefined and a client sends `$condition`, this falls through to the `rawConstraints`/no-op path and the write proceeds **unconditionally** — the `$condition` is silently dropped. A config typo (`adjustmentCondition` vs `adjustmentConditions`) or a client expecting a guard gets an unguarded delta. `updateEntity` throws `INVALID_CONDITION` for the mirror case (`no updateConditions defined`, L245-249). Make adjust symmetric: `else if (condition) throw new StandardError(StandardErrorCode.INVALID_CONDITION, ...)` before the legacy branch.

packages/core/services/entity.service.ts
```diff
@@ -104,17 +110,51 @@ export class EntityService {
...
+      console.warn(
+        '[monorise] adjustmentConstraints is deprecated. Use adjustmentConditions instead.',
```
yeuharn96, line 163: 🔵 nit: this and the `$where` warn at L262 fire on every request that hits the legacy path — noisy in prod logs at volume. Consider a module-level `Set` / once-guard so each deprecation logs once per process.

yeuharn96 (changes requested):
**🤖 Reviewed by Claude**

Solid direction — moving raw operators server-side and giving conditions names is the right call, and the test coverage is good. Three things to resolve before merge, plus a nit.

**1. The security goal isn't actually met while legacy `$where` stays client-facing.** The PR's stated Problem 1 is that DynamoDB operators must never reach frontend clients (field probing via status codes). But `updateEntity` still accepts raw `$where` from the request body, unchanged, gated by nothing but a deprecation `console.warn`. Until `$where` is removed or gated behind a server opt-in, the exact hole this PR is meant to close remains open. See inline.

**2. Static conditions check the pre-adjustment value**, so max/cap-style conditions are off by the delta and can be exceeded. The `deposit` examples demonstrate the bug. See inline.

**3. `adjustEntity` silently ignores `$condition`** when the entity has no `adjustmentConditions`, while `updateEntity` throws `INVALID_CONDITION` for the analogous case. Inconsistent, and a footgun. See inline.

nit: deprecation warnings fire on every request — consider once-per-process.

packages/core/services/entity.service.ts
wengkhing: Fixed — `$where` on `updateEntity` is now disabled by default. Added `allowLegacyWhere?: boolean` to `MonoriseEntityConfig` (opt-in per entity); without it, `$where` now throws `INVALID_CONDITION` (400) instead of silently proceeding with a `console.warn`. This closes the exact oracle you described — an entity without the flag can no longer be probed via $where. Test fixtures (WALLET/USER) opted in so the existing $where mechanics suite keeps validating real behavior against real DynamoDB via localstack; added a new test proving the secure default (COURSE, no opt-in → 400) and one proving the opt-in path still works.

examples/ledger/monorise/configs/wallet.ts
wengkhing: Fixed — this was already patched on a downstream branch (feat/transactional-writes / PR #235) but landed in the wrong layer; moved it down here where the bug actually lives. `deposit` is now delta-aware (`balance: { $lte: 1000000 - (adjustments.balance ?? 0) }`), same fix applied to `test-utils.ts`'s mock config and the `www/docs/react.md` example. Also added the boundary test you asked for: balance under the cap but delta pushes it over → 409, and landing exactly at the cap → 200 (`ConditionalAdjustEntityHttp.test.ts`).

packages/core/services/entity.service.ts
wengkhing: Fixed — added an `else if (condition)` branch before the legacy `rawConstraints` path: if the client sends `$condition` but the entity has no `adjustmentConditions` defined, it now throws `INVALID_CONDITION` (400), mirroring `updateEntity`'s handling of the same case. Added a test locking this in.

packages/core/services/entity.service.ts
wengkhing: Fixed — both warnings (`adjustmentConstraints`, `$where`) now go through a small `deprecationWarnOnce(key, message)` helper backed by a module-level `Set`, so each fires once per process instead of on every request.

wengkhing:
Rebased onto main (resolved conflicts — createMutualConfig from #240, which merged in the meantime, plus an unrelated `ttl` docs addition) and pushed 007ce5d + a follow-up addressing all 3 blockers + the nit from the review:

- **Security (`$where` still client-facing)**: added `allowLegacyWhere?: boolean` to `MonoriseEntityConfig`, default off. Without it, `updateEntity` now throws `INVALID_CONDITION` (400) on raw `$where` instead of a warn-and-proceed — the field-probing oracle is closed by default. Test fixtures opt in where the suite needs to exercise `$where` mechanics; added tests for both the secure default and the opt-in path.
- **Off-by-delta on `deposit`**: turned out this was already fixed on `feat/transactional-writes` (PR #235) — moved down here since that's where the bug actually lives, and #235 no longer carries it. Added the boundary tests you asked for (delta pushes over cap even though pre-adjustment balance is under it; landing exactly at cap).
- **Asymmetric `$condition` handling**: `adjustEntity` now throws `INVALID_CONDITION` when `$condition` is sent to an entity with no `adjustmentConditions`, matching `updateEntity`. Added a test.
- **nit**: deprecation warnings now fire once per process via a small `deprecationWarnOnce` helper.

Full suite (130 tests, real localstack) passes on both `feat/unified-conditions` and `feat/transactional-writes` after the rebase.

.changeset/unified-conditions.md
```diff
@@ -0,0 +1,13 @@
...
+'@monorise/react': minor
+'monorise': minor
+---
+
+Add named conditions system for conditional entity writes
+
+- `adjustmentConditions`: server-defined preconditions for `adjustEntity`. `$condition` required when defined. Condition functions receive `(data, adjustments)`.
+- `updateConditions`: server-defined preconditions for `updateEntity`. `$condition` always optional. Condition functions receive `(data)`.
+- Clients send a condition name (`$condition: 'withdraw'`), server resolves to DynamoDB ConditionExpression. Raw operators never exposed to frontend.
+- Deprecates `adjustmentConstraints` (backward compatible) and raw `$where` on updateEntity (backward compatible with warning).
```
yeuharn96: 🟡 risk: stale release note — `$where` is no longer "backward compatible with warning"; since 007ce5d it is **disabled by default** and returns 400 `INVALID_CONDITION` unless the entity sets `allowLegacyWhere: true`. This changeset is what consumers read when upgrading, and as written it tells them their existing `$where` calls will keep working. Reword, e.g.: "**Breaking (security):** raw `$where` on updateEntity is now rejected by default; opt in per entity with `allowLegacyWhere: true` (not recommended) or migrate to named `updateConditions`." Given the changed default, consider whether `@monorise/core` should be `major` rather than `minor`.

yeuharn96 (changes requested):
All four previous findings are resolved — verified the fixes and their regression tests (delta-aware deposit cap + exact-cap boundary test, symmetric `INVALID_CONDITION` on adjustEntity, `allowLegacyWhere` default-off gate with secure-default and opt-in tests, once-per-process deprecation warnings). CI is green.

One remaining item, introduced by the fix itself: the changeset now misdocuments the `$where` change. It still says "backward compatible with warning", but `$where` is now rejected by default — that's the release note upgrading users will read before their `$where` calls start returning 400. One-line fix, then this is good to go; I'll leave approval to a human.

.changeset/unified-conditions.md
wengkhing: Fixed — reworded the $where bullet as a breaking change with the recommended wording, and bumped `@monorise/core` and `monorise` to major (they carry the actual runtime behavior change; `@monorise/base`'s change is an additive optional field so left at minor).

wengkhing: Reverted to minor — no consumers depend on this package yet, so there's nothing external the major bump would be protecting. Kept the corrected wording (still accurately describes the runtime behavior as a breaking change for whoever adopts it), just versioned as minor.

.changeset/unified-conditions.md
```diff
@@ -0,0 +1,14 @@
+---
+'@monorise/base': minor
```
yeuharn96, line 2: ❓ q (semver): the bump was set to `major` in 8e4530e then reverted to `minor` in 00085f3 with "no consumers yet" — but `@monorise/core` is at 4.2.0 and `@monorise/react` at 6.0.1, i.e. published with release history. A consumer pinned `^4` who runs `npm update` picks up the `$where`→400 behavior break silently, which is exactly what a major bump prevents. If there genuinely are no external consumers this is fine and your call; if there are, this should be `major` for core/monorise. Flagging, not blocking.

yeuharn96 (commented):
Release note is fixed — the changeset now clearly flags the `$where` default-off change as **Breaking (security)** and documents the `allowLegacyWhere` opt-in. That resolves my blocking finding.

All earlier findings remain addressed and CI is green. One non-blocking judgment call left for you (below); I'm not approving per your workflow — that's yours to do.

yeuharn96 (approved):

PR #235, opened 2026-04-20
Transaction #2 - feat: transactional writes for atomic multi-entity operations

wengkhing:
## Summary

Monorise owns the DynamoDB key schema, marshalling, and validation. Without framework-level transaction support, users must either:
- Do separate calls and hope (data corruption risk on partial failure)
- Drop to raw DynamoDB SDK (bypassing monorise's key patterns, validation, marshalling)
- Build their own transaction layer (duplicating framework internals)

DynamoDB supports `TransactWriteItems` for atomic multi-item writes. This PR exposes it as a first-class monorise primitive, keeping schema knowledge encapsulated.

### What's included

- **`POST /core/transaction`** endpoint for atomic multi-entity operations
- **`TransactionService`** — validates, builds TransactWriteItems, executes, publishes events post-commit
- **`transactional` builder** — ergonomic, type-safe operation constructors exported from both `@monorise/core` and `@monorise/react`
- **Operations**: `createEntity`, `updateEntity`, `adjustEntity`, `deleteEntity`
- **Atomicity**: single `TransactWriteItems` call — all succeed or all fail
- **Condition support**: `adjustmentConditions` and `updateConditions` work within transactions via `$condition`
- **Event safety**: events (`ENTITY_CREATED/UPDATED/DELETED`) published only after commit succeeds
- **React SDK**: `transaction()` function for frontend usage
- **DynamoDB limit**: validates ≤100 items before execution

### Usage

```typescript
import { transaction, transactional } from 'monorise/react';
const tx = transactional; // optional alias

await transaction([
  tx.createEntity(Entity.ORDER, { customerId: '...', total: 5000 }),
  tx.adjustEntity(Entity.WALLET, walletId, { balance: -5000, $condition: 'withdraw' }),
  tx.updateEntity(Entity.POST, postId, { status: 'published', $condition: 'publish' }),
  tx.deleteEntity(Entity.PRODUCT, productId),
]);
// If withdraw condition fails → order not created, wallet untouched
```

Backend usage:
```typescript
import { transactional } from 'monorise/core';

await transactionService.executeTransaction([
  transactional.createEntity('order', { total: 5000 }),
  transactional.adjustEntity('wallet', walletId, { balance: -5000, $condition: 'withdraw' }),
]);
```

### Processor compatibility

- **Replication processor**: works via DynamoDB streams — transactional writes trigger stream events normally
- **Tag processor**: receives `ENTITY_CREATED/UPDATED` events post-commit
- **Mutual processor**: receives `ENTITY_MUTUAL_TO_CREATE/UPDATE` events post-commit
- All verified via integration tests

### Limitations

- Max 100 DynamoDB items per transaction (`createEntity` uses 2+ items, others use 1)
- `updateEntity` cannot modify unique fields within transactions (each unique field change adds 2+ items)
- No mutual operations in transactions (future scope)

## Test plan

- [x] 130 tests passing (8 test files), 29 new tests
- [x] **TransactionHttp.test.ts** (20 tests): validation, single ops, multi-ops, atomicity rollback, conditions, events, unique fields, edge cases
- [x] **TransactionProcessors.test.ts** (9 tests): replication via DynamoDB streams (create/update/delete), event shape verification for tag/mutual processors, multi-op events, no events on failure
- [x] Build succeeds across all packages
- [x] Existing 101 tests unaffected


packages/core/controllers/transaction/execute-transaction.controller.ts
```diff
@@ -0,0 +1,68 @@
...
+import { StandardError, StandardErrorCode } from '../../errors/standard-error';
+import type { TransactionService } from '../../services/transaction.service';
+
+export class ExecuteTransactionController {
+  constructor(private transactionService: TransactionService) {}
+
+  // biome-ignore lint/suspicious/noExplicitAny: Hono createMiddleware requires consistent return types
+  controller = createMiddleware(async (c): Promise<any> => {
+    const accountId = c.req.header('account-id') || '';
+    const body = await c.req.json();
```
yeuharn96: 🟡 risk: `await c.req.json()` is outside the try/catch. A malformed JSON body throws here → Hono returns 500 instead of a 400 `API_VALIDATION_ERROR`. Move it inside the try (or guard it).

packages/core/services/transaction.service.ts
```diff
@@ -0,0 +1,582 @@
...
+      );
+    }
+
+    // Phase 3: Execute transaction
+    try {
+      await this.dynamodbClient.transactWriteItems({
+        TransactItems: allTransactItems,
+      });
+    } catch (err) {
+      if (err instanceof TransactionCanceledException) {
```
yeuharn96, line 98: 🟡 risk: only `TransactionCanceledException` is mapped to a domain error. Other DynamoDB failures fall through to `throw err` → 500 in the controller: `ValidationException` for duplicate keys in one transaction (the `duplicate entityId` test's 409 only holds on DynamoDB Local — real AWS raises `ValidationException` for two ops on one item), item-size limits, and a `getEntity` NOT_FOUND thrown from `getEntityData` during condition resolution. Map `ValidationException` (and the not-found case) to a 4xx `StandardError`.

packages/core/services/transaction.service.ts
```diff
@@ -0,0 +1,582 @@
...
+          entry.data = entity.data as Record<string, unknown>;
+        } catch {
+          // Entity read failed — return without data
+        }
+      }
+    });
+    await Promise.all(readPromises);
+
+    // Phase 5: Publish events (fire-and-forget, same pattern as existing operations)
+    await Promise.allSettled(pendingEvents.map((ev) => this.publishEvent(ev)));
```
yeuharn96: 🟡 risk: `Promise.allSettled` silently swallows publish failures for all events, including the main `ENTITY_CREATED/UPDATED/DELETED`. `entity.service` `await`s the primary entity event so a failure surfaces. Replication is safe via streams, but a dropped event means tag/mutual processors never run with no signal. At minimum log rejected results.

packages/core/services/transaction.service.ts
```diff
@@ -0,0 +1,582 @@
+import type {
+  AttributeValue,
+  DynamoDB,
+  TransactWriteItem,
+} from '@aws-sdk/client-dynamodb';
+import { TransactionCanceledException } from '@aws-sdk/client-dynamodb';
+import { marshall } from '@aws-sdk/util-dynamodb';
```
yeuharn96: 🔵 nit: `marshall` (L7) and `z` (L16) are imported but unused.

packages/core/services/transaction.service.ts
```diff
@@ -0,0 +1,582 @@
...
+        'Transaction must contain at least one operation',
+      );
+    }
+
+    const allTransactItems: TransactWriteItem[] = [];
+    const pendingEvents: PendingEvent[] = [];
+    const resultEntries: TransactionResultEntry[] = [];
+
+    // Phase 1: Validate and build TransactWriteItems for each operation
+    for (const op of operations) {
```
yeuharn96: 🔵 nit: this `for` loop `await`s each op's build serially, and condition resolution does a `getEntity` read per op — N sequential round-trips before the commit. Consider `Promise.all` over the operations since builds are independent.

packages/core/services/transaction.service.ts
```diff
@@ -0,0 +1,582 @@
...
+        StandardErrorCode.INVALID_ENTITY_TYPE,
+        `Unknown entity type: '${op.entityType}'`,
+      );
+    }
+
+    // Validate all adjustment values are finite numbers
+    for (const [key, value] of Object.entries(op.adjustments)) {
+      if (typeof value !== 'number' || !Number.isFinite(value)) {
+        throw new StandardError(
+          StandardErrorCode.INVALID_ENTITY_TYPE,
```
yeuharn96: 🔵 nit: `INVALID_ENTITY_TYPE` for a non-numeric adjustment value is misleading — it's a payload validation error, not an entity-type problem. Same for the `Unknown operation` default case. Consider `API_VALIDATION_ERROR`/a dedicated code.

packages/core/types/transaction.ts
```diff
@@ -0,0 +1,49 @@
...
+  entityId?: string;
+  payload: EntitySchemaMap[T];
+};
+
+export type TransactionUpdateEntity<T extends EntityType = EntityType> = {
+  operation: 'updateEntity';
+  entityType: T;
+  entityId: string;
+  payload: Partial<EntitySchemaMap[T]>;
+  accountId?: string;
```
yeuharn96: 🔵 nit: `accountId?` on `TransactionUpdateEntity` (L15) and `TransactionAdjustEntity` (L25) is never read — accountId comes from the `account-id` header in the service. Drop it to avoid implying per-op override works.

yeuharn96 (changes requested):
**🤖 Reviewed by Claude** — targets `feat/unified-conditions`, not `main`.

Solid, well-tested feature: clean phase separation (validate → build → commit → post-commit read → publish), atomicity via a single `TransactWriteItems`, post-commit event publishing, and good coverage (rollback, conditions, unique fields, processors). A few things to address before merge — the error-mapping gap (#1) is the one most likely to bite in prod.

### Blocking / risk
- **Error mapping only handles `TransactionCanceledException`.** Every other DynamoDB error (notably the `ValidationException` raised for duplicate keys in one transaction, item-size, and a `getEntity` NOT_FOUND thrown during condition resolution) falls through to `throw err` → 500. The `should fail when creating duplicate entityId` test asserts 409, which only holds on DynamoDB Local — real AWS raises `ValidationException` for two ops on the same item, so that path is a 500 in prod.
- **Malformed JSON body → 500.** `await c.req.json()` runs before the try/catch.
- **All post-commit event publishes are swallowed by `Promise.allSettled`.** Unlike `entity.service` which `await`s the main entity event, here a failed publish of `ENTITY_CREATED/UPDATED/DELETED` is silently dropped (replication is safe via streams, but tag/mutual processors would never fire with no error surfaced).

### Nits
- Unused imports `marshall` and `z` in `transaction.service.ts`.
- Sequential per-op `getEntity` reads during condition resolution (Phase 1 loop).
- `INVALID_ENTITY_TYPE` reused for non-numeric adjustment values / unknown operation — misleading code.
- Dead `accountId?` field on the update/adjust operation types (accountId comes from the header).

packages/core/services/transaction.service.ts
wengkhing: Fixed — added `ValidationException` handling (real AWS's response for duplicate-item-in-one-transaction and similar request-shape issues) alongside `TransactionCanceledException`, mapped to a new `TRANSACTION_VALIDATION_ERROR` code -> 400. Also added a case for `ENTITY_IS_UNDEFINED`, thrown when a function-based condition's `getEntityData()` reads a nonexistent entity during Phase 1 -> 404, same as `adjust-entity.controller.ts` does for the non-transactional path. Both are locked in with tests: one mocks `transactWriteItems` to reject with a `ValidationException`-shaped error, the other exercises the real nonexistent-entity-condition path against localstack.

packages/core/controllers/transaction/execute-transaction.controller.ts
wengkhing: Fixed — moved `c.req.json()` inside the try/catch and added an explicit `err instanceof SyntaxError` branch -> 400 `API_VALIDATION_ERROR`. Test added with a raw malformed-JSON body.

packages/core/services/transaction.service.ts
wengkhing: Fixed — collect `Promise.allSettled` results and log any rejected ones (event + payload + reason) via `console.error`. Didn't switch to awaiting the main event specifically since a transaction can touch several entities with no single "primary" event to distinguish — logging every rejection seemed like the right minimum bar here rather than picking one op's event to treat differently.

packages/core/services/transaction.service.ts
wengkhing: Fixed — removed both.

packages/core/services/transaction.service.ts
wengkhing: Fixed — refactored Phase 1 into a `processOperation(op, accountId)` helper and run it via `Promise.all(operations.map(...))` instead of a sequential `for` loop. Order is preserved (Promise.all keeps result order regardless of resolution timing), so this only changes wall-clock time, not behavior.

packages/core/services/transaction.service.ts
wengkhing: Fixed — added a dedicated `TRANSACTION_INVALID_OPERATION` code, used for both the non-numeric adjustment value and the unknown-operation default case.

packages/core/types/transaction.ts
wengkhing: Fixed — dropped from both `TransactionUpdateEntity` and `TransactionAdjustEntity`. Confirmed nothing reads `.accountId` off these op types anywhere in the service, and the `transactional()` builder never sets it.

wengkhing:
Pushed fef5e9d addressing all 3 blocking risks + all 4 nits:

- **Error mapping gap**: added `ValidationException` handling (real AWS's response for duplicate-item-in-one-transaction and similar request-shape issues, distinct from `TransactionCanceledException`) and `ENTITY_IS_UNDEFINED` handling (thrown when a function-based condition reads a nonexistent entity during Phase 1). New `TRANSACTION_VALIDATION_ERROR` code -> 400; `ENTITY_IS_UNDEFINED` -> 404 matching `adjust-entity.controller.ts`'s existing convention. Locked in with a mocked-`transactWriteItems` test (can't reproduce real AWS's ValidationException against localstack) and a real nonexistent-entity-condition test.
- **Malformed JSON body**: `c.req.json()` moved inside the try/catch, explicit `SyntaxError` -> 400.
- **Swallowed event failures**: `Promise.allSettled` results are now inspected and any rejection is logged with the event/payload/reason.
- **Nits**: removed unused `marshall`/`z` imports; Phase 1 now runs via `Promise.all` over a `processOperation` helper instead of a sequential loop (order preserved); added a dedicated `TRANSACTION_INVALID_OPERATION` code instead of reusing `INVALID_ENTITY_TYPE`; dropped the dead `accountId?` field from the update/adjust operation types.

Full suite (164 tests, real localstack) passes on both `feat/unified-conditions` and `feat/transactional-writes`.

yeuharn96 (approved):

PR #240, opened 2026-04-24
feat: add createMutualConfig for centralized mutualData validation

wengkhing:
## Summary

- Add `createMutualConfig` — a new utility to define a Zod schema for `mutualData` that validates on create/update operations
- Schema is defined once and referenced from both entity configs, eliminating duplication for bidirectional relationships
- Replace the existing `z.record(z.string(), z.any())` (no-op validation) in `MutualService` with config-driven schema lookup
- Validate `mutualDataProcessor` output in `mutual-processor.ts` against the schema
- Remove stale TODO and commented-out mutual code in `createEntity` (already handled by lifecycle hook)

## Why

Currently `mutualData` accepts any shape with zero validation. Developers have to read specific mutation code to understand the expected payload. This makes DX poor and error-prone.

`createMutualConfig` provides a single source of truth for mutual data shape — defined once, validated at every entry point (direct API calls and processor output).

## Test plan

- [x] Unit tests for `createMutualConfig`, schema resolution, and validation in `MutualService`
- [x] `npm run build` passes
- [ ] Verify existing mutual create/update flows still work (backward compatible — configs without `mutualDataSchema` fall back to permissive validation)


packages/core/services/__tests__/mutual-data-schema.test.ts
```diff
@@ -0,0 +1,244 @@
...
+      mockLifecycle as any,
+    );
+
+    // Invalid payload should throw ZodError
+    await expect(
+      service.createMutual({
+        byEntityType: TestEntity.STUDENT as unknown as EntityType,
+        byEntityId: 'student-1',
+        entityType: TestEntity.COURSE as unknown as EntityType,
+        entityId: 'course-1',
```
jyungtong, line 169: i noticed we have the reversed entityType check too, let's add that to test.

packages/core/services/__tests__/mutual-data-schema.test.ts
jyungtong, line 1: Will having integration tests more valuable?

wengkhing:
@jyungtong **Re: reversed entity pair tests** — Added in latest commit. Both reject (invalid) and accept (valid) cases covered for COURSE→STUDENT direction.

**Re: integration tests** — Unit tests cover schema resolution and validation logic (including reversed entity pairs). Integration tests would need full SQS/DynamoDB infra — better covered via dev branch deployment. Created a dev-targeting PR for that purpose.

packages/core/controllers/mutual/create-mutual.controller.ts
```diff
@@ -35,8 +34,8 @@ export class CreateMutualController {
       });
 
       return c.json(mutual);
-    } catch (err) {
-      if (err instanceof ZodError) {
+    } catch (err: any) {
+      if (err?.constructor?.name === 'ZodError') {
```
yeuharn96: 🟡 risk: `err?.constructor?.name === 'ZodError'` is fragile — breaks under class-name minification and silently stops catching after a zod internals rename. If the intent was to survive duplicate zod instances, import `ZodError` and use `instanceof`, or check `err?.name === 'ZodError'` at minimum. Why the switch from `instanceof`?

packages/core/controllers/mutual/create-mutual.controller.ts
```diff
@@ -45,27 +44,18 @@ export class CreateMutualController {
...
-        err.code === StandardErrorCode.ENTITY_IS_UNDEFINED
-      ) {
-        c.status(httpStatus.BAD_REQUEST);
-        return c.json({
-          ...err.toJSON(),
-        });
-      }
-
-      throw err;
+      c.status(httpStatus.INTERNAL_SERVER_ERROR);
```
yeuharn96: 🔴 bug: this replaces `throw err` with a 500 that returns the raw `err.message` to the client, bypassing `generalErrorHandler` (no `errorId`, no centralized logging, leaks internal messages). Restore `throw err` for unknown errors and let the middleware handle them.

packages/core/controllers/mutual/update-mutual.controller.ts
```diff
@@ -34,8 +33,8 @@ export class UpdateMutualController {
       });
 
       return c.json(mutual);
-    } catch (err) {
-      if (err instanceof ZodError) {
+    } catch (err: any) {
+      if (err?.constructor?.name === 'ZodError') {
```
yeuharn96: 🟡 risk: same fragile `err?.constructor?.name === 'ZodError'` check as `create-mutual.controller.ts` — use `instanceof ZodError` or `err?.name`.

packages/core/controllers/mutual/update-mutual.controller.ts
```diff
@@ -44,17 +43,18 @@ export class UpdateMutualController {
...
-      ) {
+      if (err instanceof StandardError) {
         c.status(httpStatus.BAD_REQUEST);
         return c.json({
           ...err.toJSON(),
         });
       }
 
-      throw err;
+      c.status(httpStatus.INTERNAL_SERVER_ERROR);
```
yeuharn96: 🔴 bug: same regression — `throw err` replaced by a raw-message 500 that bypasses `generalErrorHandler`. Restore the re-throw.

yeuharn96 (dismissed):
The `createMutualConfig` feature itself is solid — single source of truth, dual-direction schema resolution, codegen + duplicate detection, docs and tests all read well. But the PR is not mergeable as-is: it carries a full debugging session (30+ `[MONORISE_DEBUG]` logs across 3 files), an unrelated error-handling rewrite that regresses behavior, and 3 committed build artifacts.

**Blockers (not inline-able):**

- 🔴 `packages/monorise/monorise-1.1.0-dev.{2,4,7}.tgz` — three build tarballs committed by accident. Remove them and add `*.tgz` to `.gitignore`.
- 🟡 `create/update-mutual.controller.ts`: broadening `err instanceof StandardError` to a blanket `→ 400` means *every* `StandardError` now returns `BAD_REQUEST`, including server-side ones (e.g. `CONDITIONAL_CHECK_FAILED`, transaction failures) that were previously re-thrown. That mislabels 500-class failures as client errors. Keep the per-code branches, or map by a server/client classification on the error.
- 🔵 `MutualConfig.entities` is declared but never used — schema resolution walks `mutualFields` instead. Either validate against `entities` or drop the field to avoid it drifting out of sync with the configs.

Inline comments below on the debug logging and the controller error path. Please strip the debug scaffolding and revert the unrelated error-handler changes so the diff is just the feature.

yeuharn96:
Correction: retracting the three `[MONORISE_DEBUG]` logging findings — confirmed these logs are intentional. The remaining requested changes stand: the committed `.tgz` artifacts, the controller error-path regression (raw-message 500 bypassing `generalErrorHandler`), the blanket `StandardError → 400` broadening, the fragile `constructor.name === 'ZodError'` check, and the unused `MutualConfig.entities` field.

One note (non-blocking): those logs print the full `mutualPayload` and entity `data` unconditionally in a published package — worth gating behind a debug flag if downstream deployments are a concern.

packages/core/controllers/mutual/create-mutual.controller.ts
wengkhing: Fixed — restored `instanceof ZodError` (re-imported `ZodError` from `zod`).

packages/core/controllers/mutual/create-mutual.controller.ts
wengkhing: Fixed — restored `throw err` for unmapped errors so `generalErrorHandler` handles them (errorId, publishErrorEvent, no raw message leak). Per-code branching restored too (`MUTUAL_EXISTS`, `ENTITY_IS_UNDEFINED` -> 400, everything else re-thrown).

packages/core/controllers/mutual/update-mutual.controller.ts
wengkhing: Fixed — same as create-mutual.controller.ts, restored `instanceof ZodError`.

packages/core/controllers/mutual/update-mutual.controller.ts
wengkhing: Fixed — restored `throw err` for unmapped errors, and per-code branching (`MUTUAL_NOT_FOUND` -> 400, everything else re-thrown).

wengkhing:
Pushed 00eb114 addressing the remaining blockers from #4700690309 / #4976688729:

- **`.tgz` artifacts**: removed the 3 committed tarballs, added `*.tgz` to `.gitignore`.
- **Controller error-path regression**: restored `throw err` for unmapped errors in both `create-mutual.controller.ts` and `update-mutual.controller.ts`, so `generalErrorHandler` handles them (errorId, `publishErrorEvent`, no raw message leak) — replied inline on each.
- **Blanket `StandardError → 400`**: restored per-code branching (create: `MUTUAL_EXISTS`/`ENTITY_IS_UNDEFINED`; update: `MUTUAL_NOT_FOUND`), everything else re-thrown — replied inline on each.
- **Fragile `constructor.name === 'ZodError'`**: restored `instanceof ZodError` in both controllers — replied inline on each.
- **Unused `MutualConfig.entities`**: kept the field (it's part of the tested/documented public API — dropping breaks existing docs/tests) but wired it into `generate.ts` codegen as a drift check: throws if the declared `entities` pair doesn't match how the mutual is actually referenced in `mutualFields`, so it can no longer silently drift out of sync.

One thing I deliberately did *not* touch: `MUTUAL_NOT_FOUND` returns 400 here (vs. `ENTITY_NOT_FOUND`'s 404 on the entity side), and `CONDITIONAL_CHECK_FAILED` gets translated to `MUTUAL_NOT_FOUND` in `Mutual.ts` rather than surfaced as 409 like the entity controller does. Both predate this PR (present since #117) and `@monorise/core` is already published past that point, so changing them now would be a status-code break for existing consumers — flagging in case it's worth a separate, versioned follow-up rather than folding into this PR.

Debug logging findings from the earlier review pass — no action needed per the retraction, confirmed intentional.

yeuharn96 (approved):

PR #247, opened 2026-04-27
fix(base): restore export on Entity enum in bundled .d.ts

wengkhing:
## Summary

- Add post-build script (`fix-dts.cjs`) to restore `export` on the `Entity` enum declaration in the bundled `.d.ts` output

## Why

When the build tool switched from `tsc` to `tsup` in #173, the bundled DTS output changed from `export declare enum Entity {}` to `declare enum Entity {}` + re-export via `export { Entity }`. This broke `declare module` augmentation in consumer projects — TypeScript couldn't merge augmented enum members into the non-exported declaration, causing `Type 'Entity.X' is not assignable to type 'Entity'` errors.

This affected both `@monorise/base` (individual) and `monorise/base` (combined package) consumers.

## Test plan

- [x] `npm run build` passes
- [x] Verified fix via `npm pack` + install in consumer project (shapeos) — Entity assignability errors resolved
- [x] Verified `dist/index.d.ts` has `export declare enum Entity {}` and clean `export {}` statement

### DIFF
--- a/.changeset/fix-entity-enum-export.md
+++ b/.changeset/fix-entity-enum-export.md
@@ -1,0 +1,6 @@
+---
+"@monorise/base": patch
+"monorise": patch
+---
+
+Fix Entity enum `declare module` augmentation by restoring `export` on the enum declaration in bundled `.d.ts` output. The tsup DTS bundler was stripping `export` from `export declare enum Entity {}`, which broke TypeScript module augmentation for consumer projects.
--- a/packages/base/fix-dts.cjs
+++ b/packages/base/fix-dts.cjs
@@ -1,0 +1,14 @@
+// Post-build fix: tsup DTS bundling converts `export declare enum` to
+// `declare enum` + re-export, which breaks `declare module` augmentation.
+// This script restores `export` on the Entity enum declaration and removes
+// the duplicate from the re-export statement.
+const fs = require('fs');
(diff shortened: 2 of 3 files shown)


yeuharn96 (approved):

jyungtong:
do we know why we using `tsup`? 
Based on their github, it's no longer maintained

wengkhing:
@jyungtong tsup was introduced in PR #173 ("combine monorise package into one") to replace `tsc` for building packages. The reason: `tsc` produces multiple output files with internal relative imports (e.g., `import { Entity } from './types/monorise.type'`), which Turbopack couldn't resolve at runtime. tsup bundles everything into a single ESM output file, eliminating those internal imports.

That said, you're right that tsup is no longer actively maintained. Worth considering alternatives like `unbuild` or `pkgroll` in the future. But that's a separate effort — this PR just addresses the immediate DTS issue.

wengkhing:
Closing — the Entity type LSP issue was not caused by the `export` keyword on the enum declaration in the bundled `.d.ts`. The root cause was that `.monorise/` is a dot-directory excluded by TypeScript's default glob, so the `declare module` augmentation in `.monorise/config.ts` was never in the compilation scope. Fix applied in PR #203 by adding `".monorise/**/*.ts"` to the root tsconfig `include`.

PR #279, opened 2026-07-15
feat(core): add entity TTL support, hardcode expiresAt as DynamoDB TTL attribute

wengkhing:
## Summary
- Adds `ttl.processor` config to `createEntityConfig` — computes an entity's DynamoDB TTL (`expiresAt`, epoch seconds), recomputed on every create/update/upsert. Returning `undefined` leaves any existing TTL untouched on update.
- Hardcodes `expiresAt` as the DynamoDB TTL attribute in `SingleTable`/`MonoriseCore`, removing the previously user-configurable `ttl`/`tableTtl` args. The codebase already assumed `expiresAt` internally for mutual/tag locks, so a mismatched or unset value silently left TTL disabled.
- Fixes several bugs found along the way (including a self-review pass after the initial version of this PR):
  - `Mutual#expiresAt` was stored/read as an ISO string; DynamoDB TTL requires epoch seconds (type `N`). Now consistent with `Entity#expiresAt`.
  - `upsertEntity` threw a DynamoDB validation error when called with an `entityId` that had never been created (nested `data.<field>` update paths require `data` to already exist as a Map). It now delegates to `createEntity` in that case, so the new entity gets the same `LIST#`/`UNIQUE#`/`EMAIL#` replica rows, `uniqueFields` enforcement, and typed ID-collision error as any other created entity — previously fixing the throw alone would have left it silently creating an incomplete, orphaned record.
  - `computeExpiresAt`'s merge-fetch swallowed any error (not just "entity not found") as if the entity didn't exist yet. Now only the specific not-found case is treated that way.
  - `ttl.processor` was always given the current operation's timestamp as `createdAt`, never the entity's true original creation date, so a creation-relative TTL would silently drift forward on every update. Now it's the real `createdAt`.
  - `updateEntity` did up to 3 redundant `GetItem` reads for a single update when both `ttl` and a changed `uniqueFields` value were involved. Now shares one fetch and constructs the returned entity locally instead of re-fetching after the transactional write.

## Example

```ts
const config = createEntityConfig({
  name: 'session',
  displayName: 'Session',
  baseSchema,
  ttl: {
    // fixed 30-day TTL from creation/each update
    processor: () => Math.floor(Date.now() / 1000) + 30 * 24 * 60 * 60,
  },
});
```

TTL can also be data-driven, computed from the entity's own fields:

```ts
const config = createEntityConfig({
  name: 'invite',
  displayName: 'Invite',
  baseSchema,
  ttl: {
    processor: (entity) => {
      return entity.data.expiresOn
        ? new Date(entity.data.expiresOn)
        : undefined; // no expiresOn set → no TTL for this record
    },
  },
});
```

`processor` is called again on every update/upsert using the entity's data as it will be *after* the update — so a sliding-expiration TTL (e.g. extending a session on activity) works out of the box.

## Test plan
- [x] `packages/core` test suite passes against localstack (99/99 tests, including new TTL, upsert-as-create, and createdAt-preservation coverage)
- [x] `@monorise/base`, `@monorise/core`, `@monorise/sst` type-check and build clean
- [x] VitePress docs site builds clean

https://claude.ai/code/session_016TQUuDjwSMFkU5HMXZ4HvG


yeuharn96 (changes requested):
🤖 **Reviewed by Claude** — Requesting changes.

Overall a strong, well-tested PR (full review in the comment above). Blocking on one substantive item:

**Issue #1 — TTL not carried onto unique-field replica rows when the processor returns `undefined`.** For an entity configured with both `uniqueFields` and `ttl.processor`, a unique-field-changing update whose processor returns `undefined` writes the new `UNIQUE#` replica row with no TTL (orphaned-replica risk once the main row is reaped) and returns an entity whose `expiresAt` disagrees with the persisted main row. It's untested because no mock entity combines the two configs.

Requesting either (a) preserving the previous `expiresAt` for replica rows on an `undefined` result, or (b) documenting the `ttl`+`uniqueFields` constraint — plus a combined-config test either way.

Issues #2 (extra GetItem per TTL update) and #3 (partial→full cast in upsert-create) are non-blocking.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

packages/core/data/Entity.ts
```diff
@@ -560,6 +667,11 @@ export class EntityRepository extends Repository {
                   new Date().toISOString(),
               },
               updatedAt: { S: entity.updatedAt || new Date().toISOString() },
+              ...(expiresAt && {
+                expiresAt: {
+                  N: String(Math.floor(expiresAt.getTime() / 1000)),
+                },
+              }),
```
yeuharn96, line 679: 🔴 bug: replica `UNIQUE#` Put only adds `expiresAt` when the `expiresAt` param is truthy. On a unique-field update where `ttl.processor` returns undefined, this fresh Put gets NO ttl while the main row keeps its old `expiresAt` (the SET expression omits it). Main row later reaped by DynamoDB → orphaned replica row. Carry the prior `expiresAt` from `previousEntity` when the processor returns undefined.

packages/core/data/Entity.ts
```diff
@@ -703,7 +847,18 @@ export class EntityRepository extends Repository {
...
+          return new Entity<T>(
+            entityType,
+            entityId,
+            { ...previousEntity.data, ...toUpdate.data },
+            previousEntity.createdAt
+              ? new Date(previousEntity.createdAt)
+              : undefined,
+            new Date(currentDatetime),
+            expiresAtDate,
+          );
```
yeuharn96, line 879: 🟡 risk: local return passes `expiresAtDate` (undefined when the processor returns undefined) → returned `expiresAt` = undefined, but the persisted main row still holds the old value. The non-unique path (L865, `ALL_NEW`) reports the retained value, so the two update paths disagree for identical input. Same root cause as L670.

packages/core/data/Entity.ts
```diff
@@ -584,9 +696,50 @@ export class EntityRepository extends Repository {
...
+
+      const hasUniqueFields = Object.keys(toUpdate.data).some((field) =>
+        uniqueFields.includes(field),
+      );
+      const hasTtlProcessor = !!this.EntityConfig[entityType]?.ttl?.processor;
+
+      // Fetched once and shared between TTL merging and uniqueFields
+      // diffing below, instead of each doing its own independent read.
+      let previousEntity: Entity<T> | undefined;
+      if (hasUniqueFields || hasTtlProcessor) {
```
yeuharn96, line 716: 🔵 nit: `hasTtlProcessor` forces a getEntity on every update of a ttl entity, even when no ttl-relevant field changed. Inherent to the merge semantics and already shared with the uniqueFields read, so fine — flagging the 0→1 read cost for ttl-without-uniqueFields entities.

packages/core/data/Entity.ts
```diff
@@ -504,16 +595,32 @@ export class EntityRepository extends Repository {
...
+        throw err;
+      }
+    }
+
+    // entity doesn't exist yet — delegate to createEntity so it gets the same
+    // LIST#/UNIQUE#/EMAIL# replica rows, uniqueFields registration, and typed
+    // ID-collision error as any other newly created entity.
+    return this.createEntity(
+      entityType,
+      payload as EntitySchemaMap[T],
```
yeuharn96, line 617: 🔵 nit: `payload as EntitySchemaMap[T]` casts Partial→full in the create fallback. Upserting a new id with a partial missing required fields creates an incomplete record, no validation error. Intended for upsert-as-create, but the cast hides it — add a comment.

packages/core/helpers/test/test-utils.ts
```diff
@@ -110,6 +111,24 @@ export const createMockEntityConfig = () => ({
       })
       .partial(),
   }),
+  [MockEntityType.SESSION]: createEntityConfig({
```
yeuharn96, line 115: ❓ q: no mock entity has both `uniqueFields` and `ttl` — USER has uniqueFields, SESSION has ttl. The Entity.ts L670 interaction is untested. Add a combined-config entity + a test covering a unique-field update where the processor returns undefined?

yeuharn96 (commented):
🤖 Reviewed by Claude — inline notes. Strong PR; one blocking correctness item (L670).

packages/core/data/Entity.ts
wengkhing: Fixed in 7f18e1c — the fresh `UNIQUE#` Put and the locally-constructed return value now both carry over `previousEntity.expiresAt` when the processor returns `undefined`, matching the main row's behavior of leaving the attribute untouched. Added a combined `uniqueFields`+`ttl` mock entity (`LICENSE`) and a test covering exactly this interaction.

packages/core/data/Entity.ts
wengkhing: Fixed in 7f18e1c as part of the same change — the return value now uses the carried-over `expiresAt` too, so it agrees with the non-unique-field update path.

packages/core/data/Entity.ts
wengkhing: Agreed, leaving as-is — this is the accepted, documented trade-off (one extra read only for ttl-configured entity types, shared with the uniqueFields read where applicable).

packages/core/data/Entity.ts
wengkhing: Added a comment in 7f18e1c explaining the cast is intentional (upsert-as-create treats the given partial payload as the complete initial record; required-field validation happens in UpsertEntityController before this is called).

packages/core/helpers/test/test-utils.ts
wengkhing: Added `MockEntityType.LICENSE` (uniqueFields + ttl combined) in 7f18e1c, with a test covering a unique-field update where the processor returns undefined.

yeuharn96 (approved):
🤖 **Reviewed by Claude** — re-review of `7f18e1c`. Blocking item resolved; approving.

**L670 (🔴 bug) — fixed.** `effectiveExpiresAtDate = expiresAtDate ?? (previousEntity.expiresAt → Date)` now carries the prior `expiresAt` onto both the fresh `UNIQUE#` replica Put and the locally-constructed return value, so main row / replica row / return value all agree when `ttl.processor` returns undefined. Epoch-seconds round-trip is exact. Scoping is correct (declared inside the `updatedUniqueFields.length > 0` block, used in both spots).

**Test gap (❓ q) — closed.** New `LICENSE` mock entity combines `uniqueFields` + `ttl`; the new test changes the unique field with `expiresInSeconds: null` (processor → undefined) and asserts the returned entity, the main row, and the new `UNIQUE#` replica row all retain the original `expiresAt`. Covers exactly the reported path.

**L612 (🔵 nit) — addressed** with a comment documenting the intentional Partial→full cast and where validation actually happens (UpsertEntityController).

L711 (🔵 nit, extra GetItem per TTL update) left as-is — inherent to merge semantics, non-blocking.

Verified locally: `@monorise/core` typechecks clean (after rebuilding `@monorise/base`, whose gitignored dist was stale). Did **not** execute the localstack integration suite here (no docker in this env) — relying on CI for the green run.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #284, opened 2026-07-17
Add global CloudWatch log retention for Monorise core

wengkhing:
## Summary
- Add `cloudwatchLogRetention` to `MonoriseCore`
- Apply the retention setting to core Lambda log groups
- Thread logging config into the table replicator function
- Document the option in the SST SDK and getting-started docs
- Add minor changeset entries for `@monorise/sst` and `monorise`

## Testing
- `rtk tsc --project packages/sst/tsconfig.json --emitDeclarationOnly --outDir packages/sst/dist`
- `npm -w @monorise/www run docs:build`

### DIFF
--- a/packages/sst/components/monorise-core.ts
+++ b/packages/sst/components/monorise-core.ts
@@ -4,12 +4,38 @@ import { createFunctionWidgets } from './dashboard';
 import { QFunction } from './q-function';
 import { SingleTable } from './single-table';
 
+type CloudWatchLogRetention =
+  | '1 day'
+  | '3 days'
+  | '5 days'
+  | '1 week'
+  | '2 weeks'
+  | '1 month'
+  | '2 months'
+  | '3 months'
+  | '4 months'
+  | '5 months'
+  | '6 months'
+  | '1 year'
+  | '13 months'
+  | '18 months'
+  | '2 years'
+  | '3 years'
+  | '5 years'
+  | '6 years'
+  | '7 years'
+  | '8 years'
+  | '9 years'
+  | '10 years'
+  | 'forever';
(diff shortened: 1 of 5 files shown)


packages/sst/components/monorise-core.ts
```diff
@@ -4,12 +4,38 @@ import { createFunctionWidgets } from './dashboard';
 import { QFunction } from './q-function';
 import { SingleTable } from './single-table';
 
+type CloudWatchLogRetention =
```
yeuharn96: 🔵 nit: `CloudWatchLogRetention` hand-duplicates SST's `keyof typeof RETENTION` (23 members). It matches exactly today, but if SST adds/removes retention values this union drifts silently. The docs already describe the type as `sst.aws.FunctionArgs['logging']['retention']` — deriving from SST's type would avoid the drift. Non-blocking, author can ignore.

_Reviewed by Claude_

yeuharn96 (approved):

packages/sst/components/monorise-core.ts
wengkhing: Addressed in 7ef9f91 — the hand-written union is gone; `CloudWatchLogRetention` is now derived from SST's own type:

```ts
type CloudWatchLogRetention = NonNullable<
  Extract<sst.aws.FunctionArgs['logging'], { retention?: unknown }>['retention']
>;
```

This resolves to SST's `Input<keyof typeof RETENTION>`, so it can't drift when SST adds/removes retention values. Verified with a type probe (accepts `'1 week'`, rejects invalid values) and `tsc --noEmit`.

PR #285, opened 2026-07-17
feat(sst): make CloudWatch dashboard toggleable

wengkhing:
## Why

SST makes it trivial to deploy many stages (test stages, per-engineer personal stages). Each stage currently creates a CloudWatch dashboard unconditionally — past the free tier that's ~$3/dashboard/month for dashboards nobody ever opens. Only long-lived environments (notably prod) actually need one.

## What

- New `cloudwatchDashboard?: { enabled?: boolean }` option on `MonoriseCore`. Set `enabled: false` to skip creating the dashboard.
- Defaults to **enabled** — fully backward compatible, existing stages see no diff on upgrade.
- Docs updated with the recommended pattern:

```ts
new monorise.module.Core('core', {
  cloudwatchDashboard: { enabled: $app.stage === 'production' },
});
```

## Notes

- Disabling on a stage where the dashboard already exists destroys it on next deploy (intended cost saving; called out in changeset + docs).
- Object shape leaves room for future options (custom name, extra widgets) without another API change.
- Changeset: minor bump for `@monorise/sst`.
- Explored via an OpenSpec change proposal locally (artifacts not committed, per repo convention).

### DIFF
--- a/.changeset/tidy-geckos-shine.md
+++ b/.changeset/tidy-geckos-shine.md
@@ -1,0 +1,5 @@
+---
+'@monorise/sst': minor
+---
+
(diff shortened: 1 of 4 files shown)


yeuharn96 (approved):
LGTM

PR #287, opened 2026-07-21
feat(analytics): add Athena data lake

wengkhing:
## Summary
- Add opt-in DynamoDB Streams -> Firehose -> S3 analytics delivery.
- Use AWS-managed S3 and Athena encryption; no customer-managed KMS key or KMS policy setup is required.
- Generate typed Athena datasets from entity and named mutual schemas.
- Expose RDS-style lowercase `snake_case` IDs: `student_id`, `talent_profile_id`, and named-mutual IDs such as `participant_id`.
- Include both mutual endpoint IDs for direct Athena joins; same-type relations use `_source_id` and `_target_id`.
- Backfill existing state and materialize idempotent Iceberg history/current tables.
- Add a separately keyed named-query API, deployment-managed Athena views, and scheduled Iceberg models for internal dashboards.

## Dashboard Usage
Define queries, views, and models in regular TypeScript modules, then pass their objects into `analytics`:

```ts
new monorise.aws.Core("Core", {
  analytics: {
    enabled: true,
    queryApi: {
      queries: {
        "daily-transaction-summary": {
          sql: `
            SELECT day, transaction_count
            FROM daily_transactions
            WHERE day >= ? AND day < ?
          `,
          parameters: {
            from: { type: "date" },
            to: { type: "date" },
          },
          resultReuse: { maxAgeMinutes: 60 },
        },
      },
    },
    views: {
      "transaction-summary-source": {
        sql: "SELECT * FROM transaction_entities",
      },
    },
    models: {
      "daily-transactions": {
        schedule: "cron(0 1 * * ? *)",
        partitionColumn: "day",
        lookbackDays: 3,
        sql: `
          SELECT CAST(occurred_at AS date) AS day, count(*) AS transaction_count
          FROM transaction_summary_source
          WHERE occurred_at >= {{windowStart}}
            AND occurred_at < {{windowEnd}}
          GROUP BY 1
        `,
      },
    },
  },
});
```

The internal Next.js server keeps the separate `ANALYTICS_API_KEYS` secret server-side and invokes a named query. The browser never receives the key or supplies SQL:

```ts
const response = await fetch(
  `${process.env.ANALYTICS_API_URL}/analytics/queries/daily-transaction-summary/executions`,
  {
    method: "POST",
    headers: {
      "content-type": "application/json",
      "x-api-key": process.env.ANALYTICS_API_KEY!,
    },
    body: JSON.stringify({ from: "2026-07-01", to: "2026-08-01" }),
  },
);
```

Poll `GET /analytics/executions/:id`, retrieve paginated rows from `GET /analytics/executions/:id/results`, or cancel through `POST /analytics/executions/:id/cancel`.

## Verification
- `npm -w @monorise/cli test`
- `npm -w @monorise/base run build`
- `npm -w @monorise/core run test -- analytics-processor.test.ts analytics-backfill-processor.test.ts analytics-materialization-processor.test.ts analytics-model-processor.test.ts`

## Follow-up
- Full DynamoDB integration suites require LocalStack at `localhost:4566`.
- The merged `main` documentation build currently fails on an unrelated dead `http://localhost:3000` link.


packages/sst/components/analytics.ts
```diff
@@ -0,0 +1,495 @@
...
+          },
+        }, { retainOnDelete: true });
+
+    this.bucket = args.resources?.bucket ?? { arn: managedBucket!.arn, name: managedBucket!.bucket };
+    this.glueDatabase = args.resources?.glueDatabase ?? { name: managedDatabase!.name };
+    this.workgroup = args.resources?.workgroup ?? { name: managedWorkgroup!.name };
+
+    if (args.queryApi?.enabled !== false && args.queryApi) {
+      validateQueries(args.queryApi.queries);
+      const queryKeys = new sst.Secret('ANALYTICS_API_KEYS', '["analytics-secret-1", "analytics-secret-2"]');
```
yeuharn96, line 234: 🔴 **bug (security): the query API ships with working credentials that are published in this repo.**

`new sst.Secret('ANALYTICS_API_KEYS', '["analytics-secret-1", "analytics-secret-2"]')` gives the secret a *fallback value*. Any deployment where the operator does not run `sst secret set ANALYTICS_API_KEYS` comes up with those two literal keys active. The route created on L241 is an `ApiGatewayV2` with no authorizer and no throttling — `x-api-key` checked inside the Lambda on L67 of `analytics-query.ts` is the only gate. So for any such deployment, anyone who has read this open-source file can execute every configured named query against the analytics data lake, and can burn unbounded Athena spend doing it. Because the default is *valid* rather than absent, there is no deploy-time or runtime signal that the deployment is unprotected.

Fix: drop the fallback entirely — `new sst.Secret('ANALYTICS_API_KEYS')` — so `sst deploy` fails until a real value is set. `apiKeys()` already fails closed when the env var is missing, so an unset secret degrades to "no access" rather than "open access".

packages/core/processors/analytics-model-processor.ts
```diff
@@ -0,0 +1,60 @@
...
+  const source = model.sql
+    .replaceAll('{{windowStart}}', `'${start.toISOString().replaceAll("'", "''")}'`)
+    .replaceAll('{{windowEnd}}', `'${end.toISOString().replaceAll("'", "''")}'`);
+  const target = `${quote(database)}.${quote(table)}`;
+  const location = `s3://${bucket}/models/${table}/`;
+  const partitionStart = start.toISOString().slice(0, 10);
+  return [
+    `CREATE TABLE IF NOT EXISTS ${target} WITH (table_type = 'ICEBERG', location = '${location}', partitioning = ARRAY['${model.partitionColumn}']) AS ${source}`,
+    `DELETE FROM ${target} WHERE ${quote(model.partitionColumn)} >= DATE '${partitionStart}'`,
+    `INSERT INTO ${target} ${source}`,
```
yeuharn96, line 52: 🔴 bug: every scheduled model run permanently drops the boundary day's early rows.

`partitionStart` (L47) is the *date* of `start`, but `{{windowStart}}` (L43) is the full *timestamp*. The `DELETE` on L50 removes the whole partition day; the `INSERT` on L51 only re-adds rows at or after the timestamp. Rows on that day before the run's clock time are deleted and never restored — and the next run's window starts a day later, so they never come back.

`analytics-model-processor.test.ts:16-17` pins the mismatch: `DELETE ... "day" >= DATE '2026-07-24'` against `INSERT ... occurred_at >= '2026-07-24T12:00:00.000Z'`. Everything on 2026-07-24 before noon is lost on that run.

Fix: floor `start` to UTC midnight before substituting, so both statements share a boundary:
```ts
const start = new Date(end.getTime() - model.lookbackDays * 86_400_000);
start.setUTCHours(0, 0, 0, 0);
```

packages/core/handles/analytics-query.ts
```diff
@@ -0,0 +1,163 @@
...
+      .catch(() => ({} as Record<string, unknown>));
+    const declared = query.parameters ?? {};
+    if (Object.keys(body).some((name) => !(name in declared))) {
+      return c.json({ message: 'Unexpected query parameter.' }, httpStatus.BAD_REQUEST);
+    }
+    const parameters = Object.entries(declared).map(([name, definition]) => {
+      const value = validParameter(body[name], definition.type);
+      if (value === undefined) throw new Error(`Invalid or missing parameter: ${name}.`);
+      return executionParameter(value, definition.type);
+    });
```
yeuharn96: 🔴 bug: invalid parameters return 500, not 400. The `.map()` runs on L87, *before* `try {` on L92, so the `Invalid or missing parameter:` throw on L89 escapes the handler. The catch on L114-116 that exists to convert it into a 400 is unreachable dead code.

Fix: move L87-91 inside the try block.

packages/core/processors/analytics-processor.ts
```diff
@@ -0,0 +1,203 @@
...
+        },
+      });
+    } catch (error) {
+      console.error('Unable to normalize analytics stream record', error);
+      failures.push({ itemIdentifier: record.dynamodb?.SequenceNumber ?? '' });
+    }
+  }
+
+  for (let index = 0; index < events.length; index += 500) {
+    const batch = events.slice(index, index + 500);
```
yeuharn96: 🟡 risk: batches are chunked by record count only. `PutRecordBatch` also caps a request at **4 MiB** total and **1000 KiB** per record. Each event carries `before` *and* `after` as stringified JSON, and DynamoDB items go to 400 KB — 500 records will exceed 4 MiB well before they exceed the count limit, and the whole call then throws.

Since the throw happens inside the loop, earlier chunks are already delivered and the retry re-sends them; dedup downstream saves correctness but the batch never drains if it is oversized on every attempt.

Fix: accumulate by encoded byte length as well, flushing at 500 records or ~4 MiB, and skip (to `failures`) any single record over 1000 KiB.

packages/core/processors/analytics-backfill-processor.ts
```diff
@@ -0,0 +1,191 @@
...
+    const response = await s3.send(new GetObjectCommand({ Bucket: record.s3.bucket.name, Key: key }));
+    if (!response.Body) throw new Error(`Backfill export object ${key} has no body.`);
+    const content = gunzipSync(await streamToBuffer(response.Body as AsyncIterable<Uint8Array>)).toString('utf8');
+    const events = content.split('\n').filter(Boolean).flatMap((line) => {
+      const attributes = JSON.parse(line).Item;
+      return attributes ? [snapshotEvent(unmarshall(attributes), config, timestamp, omittedFields)].filter(Boolean) as AnalyticsEvent[] : [];
+    });
+    for (let index = 0; index < events.length; index += 500) {
+      const batch = events.slice(index, index + 500);
+        const result = await firehose.send(new PutRecordBatchCommand({ DeliveryStreamName: deliveryStream, Records: batch.map((value) => ({ Data: Buffer.from(`${JSON.stringify(encodeAnalyticsEvent(value))}\n`) })) }));
```
yeuharn96: 🟡 risk: same 4 MiB `PutRecordBatch` cap as `analytics-processor.ts:183` — chunking is by count only. More likely to bite here: a backfill sends full entity snapshots, so 500 records × ~8 KB already reaches the limit. Share one size-aware batching helper between the two.

packages/core/processors/analytics-backfill-processor.ts
```diff
@@ -0,0 +1,191 @@
...
+
+export function snapshotEvent(
+  item: Record<string, unknown>,
+  config: Config,
+  occurredAt: Date,
+  omittedFields: Set<string>,
+): AnalyticsEvent | undefined {
+  const attributes = marshall(item, { removeUndefinedValues: true });
+  // datasetForRecord only examines key/type string attributes; retain the full item below.
+  const dataset = datasetForRecord({ dynamodb: { NewImage: attributes } }, parseAnalyticsManifest(), config);
```
yeuharn96: 🟡 risk: `parseAnalyticsManifest()` runs `JSON.parse` on `ANALYTICS_MANIFEST` for **every item in the export**, since `snapshotEvent` is called per line (L181). On a multi-million-item backfill this dominates the Lambda's CPU and can push the run past the 15-minute timeout.

Fix: parse once in `handler` and pass the manifest into `snapshotEvent`.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,294 @@
...
+    `s.${quoteIdentifier(dataset.idColumn)}`,
+    's.before_json',
+    's.after_json',
+    ...endpointNames.map((name) => `s.${name}`),
+    ...names.map((name) => `s.${name}`),
+  ].join(', ');
+
+  return [
+    `CREATE TABLE ${historyDdl} (${allColumns.join(', ')}) LOCATION ${quoteLiteral(`s3://${bucket}/curated/history/${dataset.kind === 'entity' ? 'entities' : 'mutuals'}/${dataset.name}/`)} TBLPROPERTIES ('table_type' = 'ICEBERG')`,
+    `MERGE INTO ${history} h USING (SELECT * FROM (SELECT ${selected.join(', ')}, row_number() OVER (PARTITION BY event_id ORDER BY ordering_key DESC) AS row_number FROM ${raw}) WHERE row_number = 1) s ON h.event_id = s.event_id WHEN NOT MATCHED THEN INSERT (${historyNames}) VALUES (${insertValues})`,
```
yeuharn96: 🟡 risk: the history `MERGE` reads `FROM ${raw}` with no `event_date` predicate, so every daily run scans the entire raw table. With `projection.event_date.range = '2020-01-01,NOW'` (`analytics.ts:426`) that is ~2,400 partitions today and grows forever, at Athena's per-byte-scanned pricing — for a table whose new rows are always in the last day or two.

Fix: bound the source with a lookback, e.g. `WHERE event_date >= date_format(current_date - interval '2' day, '%Y-%m-%d')`. The `WHEN NOT MATCHED` dedup on `event_id` keeps it idempotent across overlapping windows. The `current` MERGE on L176 has the same unbounded-scan shape over `history`.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,294 @@
...
+    await createTable(statements[2]);
+    await addMissingColumns(
+      database,
+      dataset.currentTable,
+      [...generatedColumns(dataset), ...dataset.columns.map(
+        (column) =>
+          `${quoteDdlIdentifier(column.name)} ${column.type === 'json' ? 'string' : column.type}`,
+      )],
+    );
+    for (const statement of [statements[1], statements[3]]) {
```
yeuharn96, line 289: 🔵 nit: `statements[0]`/`[2]`/`[1]`/`[3]` — the ordering constraint between DDL and MERGE is invisible and one reorder in `datasetSql` silently corrupts the run. Return a named object instead: `{ historyDdl, historyMerge, currentDdl, currentMerge }`.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,294 @@
...
+      await execute(statement);
+    }
+  }
```
yeuharn96, line 292: 🟡 risk: the loop is fully sequential — 6 Athena queries per dataset (2 DDL, 2 `information_schema` lookups, 2 MERGE), each polled at 1s. At ~5s per query that is ~30s per dataset, so roughly 30 datasets exhausts the 15-minute timeout on L457 of `analytics.ts` and the tail datasets are silently never materialized.

Fix: process datasets concurrently with a bounded pool (Athena's default is 20 concurrent DML queries), or fan out one invocation per dataset.

packages/sst/components/analytics.ts
```diff
@@ -0,0 +1,495 @@
...
+          { actions: ['athena:StartQueryExecution', 'athena:GetQueryExecution', 'athena:GetQueryResults'], resources: ['*'] },
+          { actions: ['glue:GetDatabase', 'glue:GetDatabases', 'glue:GetTable', 'glue:GetTables', 'glue:CreateTable', 'glue:UpdateTable'], resources: ['*'] },
+          { actions: ['s3:GetBucketLocation', 's3:GetObject', 's3:PutObject', 's3:ListBucket'], resources: [this.bucket.arn, $interpolate`${this.bucket.arn}/*`] },
+        ],
+      },
+    });
+    const initialMaterialization = new aws.lambda.Invocation(`${id}-analytics-materialization-initial`, {
+      functionName: this.schedule.nodes.function.name,
+      input: '{}',
+    }, { dependsOn: rawTables });
```
yeuharn96, line 478: 🟡 risk: `input: '{}'` is constant, so `aws.lambda.Invocation` only fires on first create. When a later deploy adds an entity or a schema field, the manifest changes but the input does not — the new Iceberg tables and columns are not created until the daily cron on L453 runs, up to 24h later. Views created on L488 `dependsOn` this invocation and will then be built against tables that do not exist yet.

The view invocation on L490 already gets this right by folding the definition into the input. Do the same here:
```ts
input: JSON.stringify({ schemaFingerprint: manifest.schemaFingerprint }),
```

packages/sst/components/single-table.ts
```diff
@@ -65,9 +67,12 @@ export class SingleTable {
               ],
             },
           },
-          stream: 'new-and-old-images',
-          ttl: 'expiresAt',
-        });
+           stream: 'new-and-old-images',
+           ttl: 'expiresAt',
+           transform: {
+             table: { pointInTimeRecovery: { enabled: true } },
```
yeuharn96: 🟡 risk: PITR is enabled for every managed table unconditionally, but analytics is opt-in (`enabled?: boolean`, "Omit or set false to leave analytics disabled"). Deployments that never turn analytics on now pay continuous PITR backup charges on the full table with no way to decline.

`SingleTable` doesn't receive the analytics args today — thread a flag through from `MonoriseCore` (which already knows on L91 of `monorise-core.ts`) and gate `pointInTimeRecovery` on it.

packages/core/handles/analytics-query.ts
```diff
@@ -0,0 +1,163 @@
...
+    : [];
+}
+
+async function ownedExecution(id: string): Promise<boolean> {
+  const item = await dynamo.send(new GetItemCommand({
+    TableName: required('ANALYTICS_EXECUTIONS_TABLE'),
+    Key: { id: { S: id } },
+    ConsistentRead: true,
+  }));
+  return Boolean(item.Item);
```
yeuharn96: 🟡 risk: `ownedExecution` only checks that the execution id exists in the table — it does not check *which* key created it. `ANALYTICS_API_KEYS` is a list, so any holder of any valid key can read the results of, or cancel, an execution started by a different key. If the list is only ever one key this is latent, but the plural name and the two-key default on `analytics.ts:234` say otherwise.

Fix: store a key identifier (a hash, not the key) alongside `queryName` on L104-111 and compare it here.

packages/sst/components/analytics.ts
```diff
@@ -0,0 +1,495 @@
...
+          ANALYTICS_API_KEYS: queryKeys.value,
+          ANALYTICS_EXECUTIONS_TABLE: executions.name,
+          ANALYTICS_QUERIES: JSON.stringify(args.queryApi.queries),
+          ANALYTICS_DATABASE: this.glueDatabase.name,
+          ANALYTICS_WORKGROUP: this.workgroup.name,
+          ANALYTICS_ATHENA_OUTPUT: $interpolate`s3://${this.bucket.name}/athena-results/`,
+        },
+        permissions: [
+          { actions: ['athena:StartQueryExecution', 'athena:GetQueryExecution', 'athena:GetQueryResults', 'athena:StopQueryExecution'], resources: ['*'] },
+          { actions: ['glue:GetDatabase', 'glue:GetDatabases', 'glue:GetTable', 'glue:GetTables'], resources: ['*'] },
```
yeuharn96, line 259: 🟡 risk: `resources: ['*']` on `athena:StartQueryExecution` lets this function run queries in **any** workgroup in the account, and the `glue:Get*` grant exposes every table in the account's catalog — not just the analytics database. This is the internet-reachable query-API Lambda, so it is the widest blast radius in the stack.

Fix: scope Athena to the workgroup ARN and Glue to the catalog/database/table ARNs for `this.glueDatabase.name`. Same for the shared `athenaPermissions` on L266-267.

packages/cli/commands/utils/analytics-manifest.ts
```diff
@@ -0,0 +1,318 @@
...
+      }
+      if (mutuals.has(mutual)) continue;
+      mutuals.add(mutual);
+      addDataset(
+        'mutual',
+        mutual.name,
+        mutual.mutualDataSchema,
+        mutual.entities ?? [config.name, field.entityType],
+      );
+    }
```
yeuharn96, line 341: 🟡 risk: two named mutuals over the same entity pair both get datasets here, but `datasetForRecord` (`analytics-processor.ts:106-122`) matches on the sorted entity-type pair alone and returns the first hit — the DynamoDB item (`MUTUAL#byType#byId#entityType`, `Mutual.ts:790`) carries no discriminator. So one dataset silently receives all the rows and the other stays permanently empty, with no error anywhere.

Fix: reject it at generation time — track seen pairs and throw when a second named mutual maps to a pair already claimed.

packages/cli/cli.ts
```diff
@@ -186,6 +194,23 @@ ${moduleAugmentations}
...
+      `Skipping unnamed mutual analytics datasets: ${manifest.unnamedMutuals.join(', ')}.`,
+    );
+  }
+  if (fs.existsSync(analyticsManifestPath)) {
+    validateSchemaEvolution(
+      JSON.parse(fs.readFileSync(analyticsManifestPath, 'utf8')) as AnalyticsManifest,
+      manifest,
+    );
+  }
+  fs.writeFileSync(analyticsManifestPath, `${JSON.stringify(manifest, null, 2)}\n`);
```
yeuharn96: 🔵 nit: identical to `generate.ts:273-287` — manifest path, unnamed-mutual warning, evolution check, and write, duplicated verbatim. Extract `writeAnalyticsManifest(monoriseOutputDir, analyticsConfigs)` into `analytics-manifest.ts` and call it from both.

packages/sst/components/single-table.ts
```diff
@@ -65,9 +67,12 @@ export class SingleTable {
...
+           },
+         });
```
yeuharn96: 🔵 nit: L70-75 picked up an extra leading space against the surrounding block — unrelated reindent noise in the diff.

yeuharn96 (changes requested):
**Reviewed by Claude** 🤖

Substantial, well-structured feature — the idempotency design in particular is careful: deterministic `event_id`s, `WHEN NOT MATCHED` dedup in the history MERGE, and conditional marker writes in `startBackfill` all hold up under retries.

Three blocking items:

- `analytics.ts:234` — the query API ships with valid default credentials that are published in this repo.
- `analytics-model-processor.ts:47-51` — every scheduled model run permanently deletes the boundary day's early rows; the existing test pins the broken behavior.
- `analytics-query.ts:87` — parameter validation throws outside the `try`, so bad input 500s and the 400 path is dead code.

The rest are scaling and blast-radius concerns worth resolving before this carries production volume: Firehose batches ignore the 4 MiB request cap, the daily materialization scans the entire raw table forever, and the initial materialization never re-runs when the schema changes.

packages/core/handles/analytics-query.ts
wengkhing: Fixed: parameter validation now runs inside the handler try/catch and returns 400.

packages/core/processors/analytics-model-processor.ts
wengkhing: Fixed by flooring `windowStart` to UTC midnight, so the DELETE and INSERT now cover identical boundaries.

packages/core/processors/analytics-backfill-processor.ts
wengkhing: Fixed with the same shared byte-aware batching helper. Oversized backfill records fail explicitly rather than issuing an invalid Firehose request.

packages/core/processors/analytics-backfill-processor.ts
wengkhing: Fixed: the backfill handler parses the manifest once and passes it into `snapshotEvent` for every exported row.

packages/core/processors/analytics-materialization-processor.ts
wengkhing: Deferred: this is a readability-only refactor with no current behavioral defect. The statements remain covered by materialization SQL tests.

packages/sst/components/analytics.ts
wengkhing: Documented the intentional local-development defaults and the required `ANALYTICS_API_KEYS` override for shared stages in Deploying and Analytics docs. Production users must set a strong secret before enabling the query API.

packages/core/processors/analytics-materialization-processor.ts
wengkhing: Deferred by design for the dedicated materialization orchestration follow-up. Risk remains for high dataset counts or large backfills because this Lambda has a 15-minute limit.

packages/sst/components/analytics.ts
wengkhing: Fixed: initial materialization input now includes the manifest schema fingerprint, so schema changes re-run table setup before dependent views.

packages/core/processors/analytics-materialization-processor.ts
wengkhing: Fixed with a three-day raw partition lookback and an affected-ID filter for current-state merges. A durable per-partition watermark remains part of the larger orchestration follow-up.

packages/core/handles/analytics-query.ts
wengkhing: Fixed: executions persist a SHA-256 fingerprint of the creating API key and status, results, and cancellation require the same fingerprint.

packages/sst/components/single-table.ts
wengkhing: Fixed: managed table PITR is now enabled only when analytics is enabled. Imported analytics tables still require PITR explicitly.

packages/cli/commands/utils/analytics-manifest.ts
wengkhing: Fixed: manifest generation now rejects a second named mutual that maps to an already claimed entity pair.

packages/sst/components/analytics.ts
wengkhing: Deferred: supplied workgroups and Glue databases expose names but not ARNs, so safely constructing least-privilege IAM ARNs would require a public resource-contract change. The query API remains limited to deployment-defined named queries; broad IAM is the residual blast-radius risk.

packages/cli/cli.ts
wengkhing: Fixed: both generators now call shared `writeAnalyticsManifest`.

packages/core/processors/analytics-processor.ts
wengkhing: Fixed with shared byte-aware batching: requests stop at 500 records or 4 MiB, and oversized stream records are returned as batch failures.

packages/sst/components/single-table.ts
wengkhing: Fixed the unrelated indentation drift.

packages/sst/components/analytics.ts
```diff
@@ -0,0 +1,497 @@
...
+          },
+        }, { retainOnDelete: true });
+
+    this.bucket = args.resources?.bucket ?? { arn: managedBucket!.arn, name: managedBucket!.bucket };
+    this.glueDatabase = args.resources?.glueDatabase ?? { name: managedDatabase!.name };
+    this.workgroup = args.resources?.workgroup ?? { name: managedWorkgroup!.name };
+
+    if (args.queryApi?.enabled !== false && args.queryApi) {
+      validateQueries(args.queryApi.queries);
+      const queryKeys = new sst.Secret('ANALYTICS_API_KEYS', '["analytics-secret-1", "analytics-secret-2"]');
```
yeuharn96, line 234: 🔴 **bug (security): still unaddressed from the last round — the query API ships with valid credentials published in this repo.**

I see the precedent: `monorise-core.ts:78` does the same thing with `new sst.Secret('API_KEYS', '["secret1", "secret2"]')`, so this is consistent with existing convention rather than an oversight. That's fair, and if you want to keep the two in sync rather than fix them piecemeal, say so and I'll drop it.

But the convention is worth breaking here specifically, because this route is a new internet-facing surface with a different risk profile: `ApiGatewayV2` with no authorizer and no throttling, where the only gate is the `x-api-key` compare on `analytics-query.ts:72`. An operator who forgets `sst secret set ANALYTICS_API_KEYS` gets a deployment that is *fully functional* — nothing fails, nothing warns — while anyone who has read this file can execute every configured named query against the data lake and drive unbounded Athena spend. A default that is valid rather than absent is what removes the signal.

Fix is one line: `new sst.Secret('ANALYTICS_API_KEYS')`. `apiKeys()` already returns `[]` and 401s everything when the env var is missing, so an unset secret degrades to "closed" instead of "open".

packages/core/processors/analytics-backfill-processor.ts
```diff
@@ -0,0 +1,218 @@
...
+      if (result.FailedPutCount) throw new Error(`Failed to deliver ${result.FailedPutCount} backfill records; retry this export object.`);
+      delivered += batch.length;
+    }
+  }
+  if (exportComplete && process.env.ANALYTICS_MATERIALIZATION_FUNCTION) {
+    // Firehose may hold the final export batch for its full buffering interval.
+    await setTimeout(310_000);
+    await lambda.send(new InvokeCommand({
+      FunctionName: process.env.ANALYTICS_MATERIALIZATION_FUNCTION,
+      InvocationType: 'Event',
```
yeuharn96, line 213: 🟡 risk: `await setTimeout(310_000)` blocks the Lambda for 5m10s to wait out Firehose's buffering interval, but it doesn't actually establish the ordering it's reaching for.

S3 delivers one notification per object, so `manifest-summary.json` arrives in its own invocation, concurrent with the sibling invocations still processing `.json.gz` data files. Sleeping 310s in *this* invocation says nothing about whether those have finished — if S3 event delivery lags or a data-file invocation retries, materialization still fires early. The 310s is tuned against `bufferingInterval: 300` on `analytics.ts:305`, which is one of the two flush triggers; `bufferingSize: 64` MiB is the other.

The good news is you no longer need it. Materialization is idempotent, and the daily cron re-runs it against a 3-day raw window (`analytics-materialization-processor.ts:115`), so backfill data is picked up within 24h regardless. The sleep buys some freshness in the common case and silently doesn't in the uncommon one.

Fix: drop the sleep and the invoke. If same-day freshness after a backfill matters, make it explicit instead — a one-shot EventBridge schedule, or a Step Functions wait state, both of which survive the invocation ending.

packages/core/processors/analytics-backfill-processor.ts
```diff
@@ -0,0 +1,218 @@
...
+    const events = content.split('\n').filter(Boolean).flatMap((line) => {
+      const attributes = JSON.parse(line).Item;
+      return attributes ? [snapshotEvent(unmarshall(attributes), config, manifest, timestamp, omittedFields)].filter(Boolean) as AnalyticsEvent[] : [];
+    });
+    const { batches, oversized } = batchAnalyticsRecords(
+      events,
+      (value) => `${JSON.stringify(encodeAnalyticsEvent(value))}\n`,
+    );
+    if (oversized.length) {
+      throw new Error(`Cannot deliver ${oversized.length} backfill records exceeding Firehose's 1000 KiB record limit.`);
```
yeuharn96, line 200: 🟡 risk: one oversized record aborts the entire export object. The throw discards the batches not yet sent, and the S3 async retry replays the whole object and hits the same record again — so a single >1000 KiB item makes that slice of the backfill permanently undeliverable, visible only as the `analytics-backfill-errors` alarm.

The stream processor handles the same case better: `analytics-processor.ts:213-215` routes oversized records to `failures` and lets the rest through. Match it — log and skip, and return the count alongside `delivered` so the drop is observable without blocking the backfill.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,295 @@
...
+  );
+  const endpointDdlNames = dataset.endpoints.map((endpoint) =>
+    quoteDdlIdentifier(endpoint.column),
+  );
+  const values = dataset.columns.map((column) =>
+    valueExpression(column, dataset.kind),
+  );
+  const payload = 'coalesce(after, before)';
+  const idPath = dataset.kind === 'entity' ? '$.entityId' : '$.mutualId';
+  const recentRaw = `${raw} WHERE event_date >= date_format(date_add('day', -3, current_date), '%Y-%m-%d')`;
```
yeuharn96, line 115: 🟡 risk: the 3-day window is the right call, but it converts a silent cron failure from self-healing into permanent loss, and nothing currently detects that failure.

The old full scan meant a broken materialization run repaired itself on the next success. Now, three consecutive failed daily runs — Athena quota, a DDL error on one dataset aborting the loop before later datasets, a deploy gap — and those events are never merged into history. There is no alarm on the schedule function: `analytics.ts` has four `MetricAlarm`s (DLQ, iterator age, Firehose freshness, backfill errors) and none covers `this.schedule.nodes.function`.

Fix: add an `Errors` alarm on the materialization function next to the one on `analytics.ts:416`. Widening the window to ~7 days is cheap insurance too — raw partitions are date-pruned, so the extra scan is small relative to the loss it prevents.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,295 @@
...
+      database,
+      dataset.currentTable,
+      [...generatedColumns(dataset), ...dataset.columns.map(
+        (column) =>
+          `${quoteDdlIdentifier(column.name)} ${column.type === 'json' ? 'string' : column.type}`,
+      )],
+    );
+    for (const statement of [statements[1], statements[3]]) {
+      await execute(statement);
+    }
```
yeuharn96, line 291: 🟡 risk: unchanged from the last round, and now it interacts with the lookback above. The loop is sequential — 6 Athena queries per dataset, polled at 1s. At ~5s per query that is ~30s per dataset, so roughly 30 datasets exhausts the 15-minute timeout on `analytics.ts:357` and the tail datasets are never materialized. With the 3-day window, three consecutive timeouts lose their events outright.

An unhandled failure on any single dataset also aborts every dataset after it, for the same reason.

Fix: bound-parallel the datasets (Athena's default is 20 concurrent DML queries), or fan out one invocation per dataset. Wrapping each dataset in a try/catch that records the failure and continues would at least stop one bad dataset from starving the rest.

packages/core/processors/analytics-materialization-processor.ts
```diff
@@ -0,0 +1,295 @@
...
+    await createTable(statements[2]);
+    await addMissingColumns(
+      database,
+      dataset.currentTable,
+      [...generatedColumns(dataset), ...dataset.columns.map(
+        (column) =>
+          `${quoteDdlIdentifier(column.name)} ${column.type === 'json' ? 'string' : column.type}`,
+      )],
+    );
+    for (const statement of [statements[1], statements[3]]) {
```
yeuharn96, line 289: 🔵 nit: unchanged — `statements[0]`/`[2]`/`[1]`/`[3]` still encodes the DDL-before-MERGE ordering as array indices. Return `{ historyDdl, historyMerge, currentDdl, currentMerge }` from `datasetSql` instead.

packages/sst/components/analytics.ts
```diff
@@ -0,0 +1,497 @@
...
+          ANALYTICS_API_KEYS: queryKeys.value,
+          ANALYTICS_EXECUTIONS_TABLE: executions.name,
+          ANALYTICS_QUERIES: JSON.stringify(args.queryApi.queries),
+          ANALYTICS_DATABASE: this.glueDatabase.name,
+          ANALYTICS_WORKGROUP: this.workgroup.name,
+          ANALYTICS_ATHENA_OUTPUT: $interpolate`s3://${this.bucket.name}/athena-results/`,
+        },
+        permissions: [
+          { actions: ['athena:StartQueryExecution', 'athena:GetQueryExecution', 'athena:GetQueryResults', 'athena:StopQueryExecution'], resources: ['*'] },
+          { actions: ['glue:GetDatabase', 'glue:GetDatabases', 'glue:GetTable', 'glue:GetTables'], resources: ['*'] },
```
yeuharn96, line 259: 🟡 risk: unchanged — `resources: ['*']` on `athena:StartQueryExecution` lets this function run queries in any workgroup in the account, and `glue:Get*` exposes the entire account catalog rather than the analytics database. This is the internet-reachable query-API Lambda, so it is the widest blast radius in the stack, and it compounds the credential default on L234.

Fix: scope Athena to the workgroup ARN and Glue to the catalog/database/table ARNs for `this.glueDatabase.name`. Same for `athenaPermissions` on L266-267 and the materialization grants on L368-369.

yeuharn96 (changes requested):
**Reviewed by Claude** 🤖

Good turnaround — 12 of 16 addressed, and several fixes went further than suggested. Verified locally on `b981c41`: `npm -w @monorise/core run test -- analytics` 9/9 and `npm -w @monorise/cli test` 8/8 green.

Confirmed fixed:
- Model refresh data loss — `start.setUTCHours(0,0,0,0)`, with the test updated to assert the aligned boundary.
- Parameter validation moved inside the `try`, so bad input 400s.
- Firehose batching is now byte-aware in both paths, with a real regression test on the split.
- Raw scan bounded — and you bounded the `current` MERGE too via the `IN (SELECT DISTINCT ...)` subquery, which I hadn't suggested.
- Execution ownership scoped by `ownerKeyHash`; PITR gated on `analyticsEnabled`; materialization invocation keyed on `schemaFingerprint`; named-mutual pair collision now rejected at generation; manifest write extracted to `writeAnalyticsManifest`.

Still blocking:
- **`analytics.ts:234`** — the default API keys are untouched. I've acknowledged the existing `API_KEYS` precedent in the inline comment; if the intent is to fix both together in a separate PR, say so and I'll drop it.

Two new items from this commit, plus three carried over:
- The 310s blocking sleep in the backfill doesn't establish the ordering it's aiming for, and the daily cron already covers the case.
- An oversized backfill record now aborts its whole export object, where the stream processor skips and continues.
- The 3-day lookback needs an alarm behind it — a silently failing materialization is now permanently lossy, and no alarm covers that function.

packages/sst/components/analytics.ts
yeuharn96: Withdrawing this — I missed your reply on the original thread before re-raising it, apologies.

You addressed it deliberately and documented it: `deploying.md` has a "Set API keys (required)" section listing `ANALYTICS_API_KEYS`, and `analytics.md:91` states outright that "the local-development default analytics keys are public." That's a legitimate answer to a design-tradeoff finding, and it's consistent with the existing `API_KEYS`/`X_API_KEY` handling rather than a one-off. Your call stands; no change needed.

packages/core/processors/analytics-materialization-processor.ts
yeuharn96: Withdrawing — you already marked this deferred as readability-only and I re-raised it without reading that. Agreed there's no behavioral defect.

packages/core/processors/analytics-materialization-processor.ts
yeuharn96: Withdrawing the re-raise — you'd already deferred this to the materialization orchestration follow-up and named the same 15-minute risk I repeated.

One thing from the second pass that is genuinely new rather than a repeat: the 3-day lookback now means three consecutive timeouts lose those events outright, where the old full scan self-healed. That's an argument for the alarm on the other thread, not for doing the orchestration work in this PR.

packages/sst/components/analytics.ts
yeuharn96: Withdrawing the re-raise — I missed your reply. Your reason is a real constraint: `SuppliedAnalyticsResources` exposes `workgroup?: { name }` and `glueDatabase?: { name }` with no ARNs, so the supplied-resource path genuinely can't build least-privilege ARNs today.

For whenever the follow-up happens: both ARNs are derivable from name plus account and region (`arn:aws:athena:<region>:<account>:workgroup/<name>`, `arn:aws:glue:<region>:<account>:database/<name>`) via `aws.getCallerIdentity()`, so it may not need the contract change. Not for this PR — just noting it so the deferral isn't based on it being impossible.

packages/core/processors/analytics-backfill-processor.ts
yeuharn96: Reframing this — your reply on the batching thread already said oversized backfill records fail explicitly by design, which I should have picked up before filing it as a new finding.

The deliberate fail-loud choice is reasonable. The part I'd still flag is narrower: the throw also discards the batches queued behind it in the same object, and the S3 async retry replays the object and hits the same record again, so the surrounding good records never land either. Skipping the oversized record and returning its count alongside `delivered` keeps the loud failure signal without taking the rest of the export with it. Your call — not blocking.

yeuharn96:
**Reviewed by Claude** 🤖 — correction to my previous review

I ran the second review off the code diff alone and did not read your replies first. Four of the seven comments I posted were re-raises of items you had already answered or explicitly deferred, and I labelled them "unchanged" and "still unaddressed" as though they'd been ignored. That was wrong and I've withdrawn each one in its thread.

Specifically withdrawn: the `ANALYTICS_API_KEYS` default (you addressed it by documenting it in `deploying.md` and `analytics.md:91`, consistent with the existing `API_KEYS` convention), the `statements[0..3]` indices, the sequential materialization loop, and the IAM `resources: ['*']` scoping. The backfill oversized-record comment is reframed — your fail-loud choice was already stated; only the "discards the batches behind it" detail is additional, and it isn't blocking.

The `CHANGES_REQUESTED` state on that review was therefore based mostly on a mistake. Two items from it are genuinely new and still open, neither clearly blocking:

- The 310s blocking sleep in the backfill — doesn't establish the ordering it's aiming for, and the daily cron plus 3-day window already covers the case.
- No `Errors` alarm on the materialization function — this one is a consequence of the lookback fix, since a silently failing run is now permanently lossy where the full scan used to self-heal.

Happy for both to be follow-ups if you'd rather land this. Leaving the approve/merge decision to @yeuharn96.

yeuharn96 (approved):

PR #292, opened 2026-07-29
fix(cli): defer custom route loading

wengkhing:
Avoid evaluating application custom routes during code generation.

Also provide a valid temporary generated config while entity modules resolve the `#/monorise` alias.

Includes patch release metadata for `@monorise/cli` and `monorise`.

### DIFF
--- a/packages/cli/cli.ts
+++ b/packages/cli/cli.ts
@@ -48,6 +48,10 @@ export enum Entity {}
 export default {};
 `;
   fs.writeFileSync(configOutputPath, initialConfigContent);
+  fs.writeFileSync(
+    path.join(monoriseOutputDir, 'index.ts'),
+    `export * from './config';\nexport { default } from './config';\n`,
+  );
 
   const files = fs
     .readdirSync(configDir)
@@ -260,67 +264,6 @@ export const appHandler = coreFactory.appHandler(${appHandlerPayload});
   return handleOutputPath;
 }
 
-const TSCONFIG_SCAN_IGNORE = new Set([
-  'node_modules',
-  '.git',
-  '.monorise',
-  '.next',
-  '.turbo',
-  '.sst',
-  'dist',
-]);
-
-function findTsconfigFiles(dir: string): string[] {
-  const results: string[] = [];
-  let entries: fs.Dirent[];
-  try {
-    entries = fs.readdirSync(dir, { withFileTypes: true });
-  } catch {
-    return results;
-  }
-
-  for (const entry of entries) {
-    if (entry.isDirectory()) {
-      if (TSCONFIG_SCAN_IGNORE.has(entry.name)) continue;
-      results.push(...findTsconfigFiles(path.join(dir, entry.name)));
-    } else if (entry.name === 'tsconfig.json') {
-      results.push(path.join(dir, entry.name));
-    }
-  }
-
-  return results;
-}
-
-// Existing projects (scaffolded before the '#/monorise' shorthand existed, or
-// upgrading @monorise/cli without re-running `init`) never get their
-// tsconfig.json touched again after scaffolding — so the exact-match alias
-// for '#/monorise' can be missing even though '#/monorise/*' already works.
-// Since this runs on every dev/build, it keeps any tsconfig that already
-// opted into the '#/monorise/*' wildcard in sync with the exact-match sibling.
-function syncMonoriseAliasInTsconfigs(projectRoot: string): void {
-  const wildcardKey = '#/monorise/*';
-  const exactKey = '#/monorise';
-
-  for (const tsconfigPath of findTsconfigFiles(projectRoot)) {
-    try {
-      const content = fs.readFileSync(tsconfigPath, 'utf8');
-      const tsconfig = JSON.parse(content);
-      const paths = tsconfig.compilerOptions?.paths;
-      const wildcardTarget = paths?.[wildcardKey]?.[0];
-
-      if (!wildcardTarget || paths[exactKey]) continue;
-
-      paths[exactKey] = [wildcardTarget.replace(/\*$/, 'index')];
-      fs.writeFileSync(tsconfigPath, JSON.stringify(tsconfig, null, 2));
-      console.log(
-        `Added '#/monorise' path alias to ${path.relative(projectRoot, tsconfigPath)}`,
-      );
-    } catch {
-      // Best-effort — a malformed or unreadable tsconfig shouldn't fail the build.
-    }
-  }
-}
-
 async function generateFiles(rootPath?: string): Promise<string> {
   const baseDir = rootPath ? path.resolve(rootPath) : process.cwd();
   const configFilePathTS = path.join(baseDir, 'monorise.config.ts');
@@ -348,7 +291,6 @@ async function generateFiles(rootPath?: string): Promise<string> {
 
   await generateConfigFile(configDir, monoriseOutputDir, projectRoot);
   await generateHandleFile(monoriseConfig, projectRoot, monoriseOutputDir);
-  syncMonoriseAliasInTsconfigs(projectRoot);
 
   return configDir;
 }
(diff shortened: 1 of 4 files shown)


packages/cli/cli.ts
```diff
@@ -45,6 +45,7 @@ async function generateConfigFile(
   const configOutputPath = path.join(monoriseOutputDir, 'config.ts');
   const initialConfigContent = `
 export enum Entity {}
+export default {};
```
yeuharn96, line 48: 🔴 bug: placeholder is incomplete — `.monorise/index.ts` is only written at L190, *after* the entity modules are imported at L75. The exact-match alias `#/monorise` maps to `./.monorise/index` (see L701, `templates/root-tsconfig.ts:13`), so on a first run or after `rm -rf .monorise` the alias still fails to resolve during the exact window this placeholder exists to cover. `#/monorise/config` works; `#/monorise` doesn't.

Write the placeholder index alongside it:
```ts
fs.writeFileSync(configOutputPath, initialConfigContent);
fs.writeFileSync(
  path.join(monoriseOutputDir, 'index.ts'),
  `export * from './config';\nexport { default } from './config';\n`,
);
```

packages/cli/cli.ts
```diff
@@ -285,6 +260,67 @@ export const appHandler = coreFactory.appHandler(${appHandlerPayload});
...
+  return results;
+}
+
+// Existing projects (scaffolded before the '#/monorise' shorthand existed, or
+// upgrading @monorise/cli without re-running `init`) never get their
+// tsconfig.json touched again after scaffolding — so the exact-match alias
+// for '#/monorise' can be missing even though '#/monorise/*' already works.
+// Since this runs on every dev/build, it keeps any tsconfig that already
+// opted into the '#/monorise/*' wildcard in sync with the exact-match sibling.
+function syncMonoriseAliasInTsconfigs(projectRoot: string): void {
```
yeuharn96: Architectural — flagging with rationale rather than a one-liner, since the fix is a design decision.

`syncMonoriseAliasInTsconfigs` silently rewrites user-owned `tsconfig.json` files on every `dev` and `build`. Three problems with that:

1. **Reformatting.** `JSON.stringify(tsconfig, null, 2)` rewrites the whole file and drops the trailing newline. Any project whose tsconfig isn't already 2-space-and-no-final-newline gets a spurious diff, and repos running biome/prettier on `tsconfig.json` will fight this on every build.
2. **JSONC silently loses.** `tsconfig.json` is JSONC by convention and comments are common. `JSON.parse` throws, the bare `catch` at L318 swallows it, and the alias is never added — for exactly the legacy projects this function targets. The user gets no signal at all.
3. **Scope.** Neither the PR title nor the body mentions touching consumer tsconfigs; a reader reviewing "defer custom route loading" won't expect a codegen step that mutates source files.

Suggested direction: detect and *warn* (`console.warn("Add '#/monorise': ['./.monorise/index'] to compilerOptions.paths in <file>")`) rather than write, and move the actual mutation into `monorise init` or an explicit opt-in flag, where file rewriting is expected. If you keep the auto-write, at minimum log the caught error instead of swallowing it, and preserve the trailing newline.

packages/monorise/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "monorise",
-  "version": "1.6.0",
+  "version": "1.6.1",
```
yeuharn96: 🔴 bug: double version bump. `.changeset/pretty-geckos-begin.md` declares `'monorise': patch`, so the release workflow will take this manual `1.6.1` to `1.6.2` and prepend a second changelog entry for the same change. `@monorise/cli` is also in the changeset but not bumped here, so the two packages are handled inconsistently.

Revert this file and `packages/monorise/CHANGELOG.md`; let `changeset version` produce both.

packages/cli/cli.ts
```diff
@@ -285,6 +260,67 @@ export const appHandler = coreFactory.appHandler(${appHandlerPayload});
...
+  let entries: fs.Dirent[];
+  try {
+    entries = fs.readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return results;
+  }
+
+  for (const entry of entries) {
+    if (entry.isDirectory()) {
+      if (TSCONFIG_SCAN_IGNORE.has(entry.name)) continue;
```
yeuharn96: 🟡 risk: unbounded recursive walk of the entire project root on every `dev`/`build`. `TSCONFIG_SCAN_IGNORE` misses `build`, `out`, `coverage`, `.cache`, `.output`, `.vercel`, `.venv` — a large monorepo pays a full tree stat on every invocation.

Scan the workspace roots from the root `package.json` `workspaces` / `pnpm-workspace.yaml` instead, or cap depth at 3.

packages/cli/cli.ts
```diff
@@ -285,6 +260,67 @@ export const appHandler = coreFactory.appHandler(${appHandlerPayload});
...
+  for (const tsconfigPath of findTsconfigFiles(projectRoot)) {
+    try {
+      const content = fs.readFileSync(tsconfigPath, 'utf8');
+      const tsconfig = JSON.parse(content);
+      const paths = tsconfig.compilerOptions?.paths;
+      const wildcardTarget = paths?.[wildcardKey]?.[0];
+
+      if (!wildcardTarget || paths[exactKey]) continue;
+
+      paths[exactKey] = [wildcardTarget.replace(/\*$/, 'index')];
```
yeuharn96: 🟡 risk: `replace(/\*$/, 'index')` is a no-op when `wildcardTarget` doesn't end in `*` (e.g. `./.monorise/*.ts`), writing `'#/monorise': ['./.monorise']` — an alias pointing at a directory. Guard first:
```ts
if (!wildcardTarget.endsWith('*')) continue;
```

packages/cli/cli.ts
```diff
@@ -223,32 +224,6 @@ async function generateHandleFile(
...
-          'post' in routesExport &&
-          'use' in routesExport
-        ))
-    ) {
-      throw new Error(
-        `Custom routes file at '${absoluteCustomRoutesPath}' must default export an instance of Hono (or an object with .get, .post, .use methods). Or a function that consume the dependency container provided by route handler.`,
-      );
-    }
-
     let relativePathToRoutes = path.relative(
```
yeuharn96, line 231: ❓ q: with the default-export validation gone, a `customRoutes` file that exports nothing (or the wrong shape) now surfaces as a type error inside generated `.monorise/handle.ts` instead of a CLI error naming the user's file. Worth keeping a cheap static check — regex the source for `export default` / `export { default` and throw the old message — so the actionable error survives without evaluating the module?

yeuharn96 (changes requested):
🤖 **Reviewed by Claude**

Deferring the custom-route import is the right call — importing app code mid-codegen was a genuine chicken-and-egg. Two blockers below (incomplete placeholder, double version bump) plus an architectural concern on the tsconfig rewriting, which isn't mentioned in the PR description.

packages/cli/cli.ts
wengkhing: Fixed in fd39dd7: the initial generated `index.ts` is now written before entity modules are imported.

packages/cli/cli.ts
wengkhing: Removed in fd39dd7. Generation no longer rewrites consumer tsconfig files.

packages/monorise/package.json
wengkhing: Fixed in fd39dd7: restored `monorise` to 1.6.0 and removed the manual changelog entry. The changeset now owns both version bumps.

packages/cli/cli.ts
wengkhing: Removed in fd39dd7 along with the consumer-tsconfig rewriter.

packages/cli/cli.ts
wengkhing: Deferred validation intentionally. A regex cannot reliably validate TypeScript default exports, and the generated runtime import preserves the module-resolution error with the original route path.

wengkhing:
Addressed the review in fd39dd7.

- Create the generated index placeholder before importing entities
- Remove consumer tsconfig rewriting
- Return both package version bumps to Changesets
- Verified a fresh ESM fixture with an entity importing `#/monorise` completes generation
- `npm run build` passes

yeuharn96 (approved):

PR #294, opened 2026-07-30
fix(react): keep transactional browser-safe

wengkhing:
## Summary
- implement the `transactional` operation builder in `@monorise/react`
- preserve the `monorise/react` public API without re-exporting `@monorise/core`
- add patch releases for `@monorise/react` and `monorise`

## Why
Re-exporting `transactional` from Core pulls AWS SDK and Node-only modules (`fs`, `async_hooks`) into frontend bundles. Next.js applications importing `monorise/react` then fail browser compilation even when they do not use `transactional`.

## Verification
- `npm run build`
- packed the generated `monorise` package and installed it in an isolated Humanthis checkout
- confirmed Backoffice Webpack compilation proceeds past the prior Node-module resolution errors

### DIFF
--- a/.changeset/bright-mice-hug.md
+++ b/.changeset/bright-mice-hug.md
@@ -4,3 +4,11 @@
 ---
 
 Keep the React transactional builder browser-safe.
+
+- `@monorise/react` no longer re-exports `transactional` from `@monorise/core`,
+  so browser bundles no longer pull in server-only Core code (AWS SDK, `fs`,
+  `async_hooks`). React exposes its own builder with the same wire format.
+- The unified `monorise` root no longer exports `transactional` (the name is
+  ambiguous between the Core and React copies). Use `monorise/core` on the
+  server or `monorise/react` in the browser instead. All other root exports
+  are unchanged.
--- a/.changeset/bright-mice-hug.md
+++ b/.changeset/bright-mice-hug.md
@@ -1,0 +1,6 @@
+---
+'@monorise/react': patch
+'monorise': patch
+---
+
+Keep the React transactional builder browser-safe.
--- a/packages/monorise/build.js
+++ b/packages/monorise/build.js
@@ -92,8 +92,14 @@ packages.forEach((pkg) => {
   }
 });
 
-// Create a root entrypoint that exposes packages without flattening their APIs.
-const mainIndexContent = `// Provide named exports for each package
+// Create the main index.js file that re-exports everything.
+const mainIndexContent = `// Re-export all packages from their respective modules
+export * from './base/index.js';
+export * from './core/index.js';
+export * from './react/index.js';
+export * from './sst/index.js';
+
+// Also provide named exports for each package
 export * as base from './base/index.js';
 export * as core from './core/index.js';
 export * as react from './react/index.js';
@@ -103,7 +109,13 @@ export * as sst from './sst/index.js';
 fs.writeFileSync(path.join(distDir, 'index.js'), mainIndexContent);
 
 // Create the main index.d.ts file
-const mainIndexDtsContent = `// Provide named exports for each package
+const mainIndexDtsContent = `// Re-export all packages from their respective modules
+export * from './base/index';
+export * from './core/index';
+export * from './react/index';
+export * from './sst/index';
+
+// Also provide named exports for each package
 export * as base from './base/index';
 export * as core from './core/index';
 export * as react from './react/index';
--- a/packages/monorise/build.js
+++ b/packages/monorise/build.js
@@ -92,14 +92,8 @@ packages.forEach((pkg) => {
   }
 });
 
-// Create the main index.js file that re-exports everything
-const mainIndexContent = `// Re-export all packages from their respective modules
-export * from './base/index.js';
-export * from './core/index.js';
-export * from './react/index.js';
-export * from './sst/index.js';
-
-// Also provide named exports for each package
+// Create a root entrypoint that exposes packages without flattening their APIs.
+const mainIndexContent = `// Provide named exports for each package
 export * as base from './base/index.js';
 export * as core from './core/index.js';
 export * as react from './react/index.js';
@@ -109,13 +103,7 @@ export * as sst from './sst/index.js';
 fs.writeFileSync(path.join(distDir, 'index.js'), mainIndexContent);
 
 // Create the main index.d.ts file
-const mainIndexDtsContent = `// Re-export all packages from their respective modules
-export * from './base/index';
-export * from './core/index';
-export * from './react/index';
-export * from './sst/index';
-
-// Also provide named exports for each package
+const mainIndexDtsContent = `// Provide named exports for each package
 export * as base from './base/index';
 export * as core from './core/index';
 export * as react from './react/index';
--- a/packages/monorise/build.js
+++ b/packages/monorise/build.js
@@ -2,7 +2,7 @@
 
 import fs from 'fs';
 import path from 'path';
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
@@ -109,27 +109,55 @@ export * as sst from './sst/index.js';
 fs.writeFileSync(path.join(distDir, 'index.js'), mainIndexContent);
 
 // Create the main index.d.ts file.
-// Core and React declare different `transactional` builders, so star-exporting
-// both would emit TS2308. Runtime star-exports already drop the ambiguous name
-// (ESM semantics); enumerate Core and React exports here minus `transactional`
-// so the type surface matches the runtime surface. Names exported by both
-// (e.g. `Mutual`) resolve to Core's, matching star-export behavior.
+// ESM star-exports drop ambiguous names from the runtime namespace (e.g.
+// `transactional`, declared by both Core and React, and `Entity`, exported as
+// different values by base and Core), while TS reports TS2308 for them. To
+// keep the type surface identical to the runtime surface, enumerate every
+// package's exports explicitly: names ambiguous at runtime are omitted, and
+// names declared by more than one package are claimed by the first package
+// listed (same-symbol re-exports from base, and type-only re-exports such as
+// React's `Mutual`, which tsup elides at runtime).
+const dtsPackages = ['core', 'base', 'react', 'sst'];
+
+const runtimeAmbiguous = async () => {
+  const counts = new Map();
+  for (const pkg of dtsPackages) {
+    const mod = await import(
+      pathToFileURL(path.join(distDir, pkg, 'index.js')).href
+    );
+    for (const name of Object.keys(mod)) {
+      if (name !== 'default') {
+        counts.set(name, (counts.get(name) ?? 0) + 1);
+      }
+    }
+  }
+  return new Set(
+    [...counts].filter(([, count]) => count > 1).map(([name]) => name),
+  );
+};
+
 const getDtsExportNames = (pkg) => {
   const dts = fs.readFileSync(path.join(distDir, pkg, 'index.d.ts'), 'utf-8');
   const names = new Set();
-  const exportClause = /export\s+(?:type\s+)?\{([^}]*)\}/g;
   let match;
+  const exportClause = /export\s+(?:type\s+)?\{([^}]*)\}/g;
   while ((match = exportClause.exec(dts)) !== null) {
     for (const part of match[1].split(',')) {
       const name = part
         .trim()
         .split(/\s+as\s+/)
         .pop()
-        .trim();
+        .trim()
+        .replace(/^type\s+/, '');
       if (name && name !== 'default') {
         names.add(name);
       }
     }
+  }
+  const exportDeclare =
+    /export\s+declare\s+(?:const|let|var|function|class|interface|type|enum|namespace)\s+([A-Za-z_$][\w$]*)/g;
+  while ((match = exportDeclare.exec(dts)) !== null) {
+    names.add(match[1]);
   }
   if (names.size === 0) {
     throw new Error(`Failed to enumerate exports from ${pkg}/index.d.ts`);
@@ -137,22 +165,26 @@ const getDtsExportNames = (pkg) => {
   return names;
 };
 
-const coreNames = getDtsExportNames('core');
-coreNames.delete('transactional');
-const reactNames = getDtsExportNames('react');
-reactNames.delete('transactional');
-for (const name of coreNames) {
-  reactNames.delete(name);
-}
-
-const toExportClause = (names, pkg) =>
-  `export {\n${[...names].map((name) => `  ${name},`).join('\n')}\n} from './${pkg}/index';`;
+const ambiguous = await runtimeAmbiguous();
+const claimed = new Set();
+const dtsClauses = dtsPackages
+  .map((pkg) => {
+    const names = [...getDtsExportNames(pkg)].filter(
+      (name) => !ambiguous.has(name) && !claimed.has(name),
+    );
+    for (const name of names) {
+      claimed.add(name);
+    }
+    if (names.length === 0) {
+      return '';
+    }
+    return `export {\n${names.map((name) => `  ${name},`).join('\n')}\n} from './${pkg}/index';`;
+  })
+  .filter(Boolean)
+  .join('\n');
 
 const mainIndexDtsContent = `// Re-export all packages from their respective modules
-export * from './base/index';
-${toExportClause(coreNames, 'core')}
-${toExportClause(reactNames, 'react')}
-export * from './sst/index';
+${dtsClauses}
 
 // Also provide named exports for each package
 export * as base from './base/index';
--- a/packages/monorise/build.js
+++ b/packages/monorise/build.js
@@ -108,11 +108,50 @@ export * as sst from './sst/index.js';
 
 fs.writeFileSync(path.join(distDir, 'index.js'), mainIndexContent);
 
-// Create the main index.d.ts file
+// Create the main index.d.ts file.
+// Core and React declare different `transactional` builders, so star-exporting
+// both would emit TS2308. Runtime star-exports already drop the ambiguous name
+// (ESM semantics); enumerate Core and React exports here minus `transactional`
+// so the type surface matches the runtime surface. Names exported by both
+// (e.g. `Mutual`) resolve to Core's, matching star-export behavior.
+const getDtsExportNames = (pkg) => {
+  const dts = fs.readFileSync(path.join(distDir, pkg, 'index.d.ts'), 'utf-8');
+  const names = new Set();
+  const exportClause = /export\s+(?:type\s+)?\{([^}]*)\}/g;
+  let match;
+  while ((match = exportClause.exec(dts)) !== null) {
+    for (const part of match[1].split(',')) {
+      const name = part
+        .trim()
+        .split(/\s+as\s+/)
+        .pop()
+        .trim();
+      if (name && name !== 'default') {
+        names.add(name);
+      }
+    }
+  }
+  if (names.size === 0) {
+    throw new Error(`Failed to enumerate exports from ${pkg}/index.d.ts`);
+  }
+  return names;
+};
+
+const coreNames = getDtsExportNames('core');
+coreNames.delete('transactional');
+const reactNames = getDtsExportNames('react');
+reactNames.delete('transactional');
+for (const name of coreNames) {
+  reactNames.delete(name);
+}
+
+const toExportClause = (names, pkg) =>
+  `export {\n${[...names].map((name) => `  ${name},`).join('\n')}\n} from './${pkg}/index';`;
+
 const mainIndexDtsContent = `// Re-export all packages from their respective modules
 export * from './base/index';
-export * from './core/index';
-export * from './react/index';
+${toExportClause(coreNames, 'core')}
+${toExportClause(reactNames, 'react')}
 export * from './sst/index';
 
 // Also provide named exports for each package
--- a/packages/react/helpers/transactional.ts
+++ b/packages/react/helpers/transactional.ts
@@ -1,13 +1,13 @@
 import type { Entity, EntitySchemaMap } from '@monorise/base';
 
-type TransactionCreateEntity<T extends Entity> = {
+export type TransactionCreateEntity<T extends Entity = Entity> = {
   operation: 'createEntity';
   entityType: T;
   entityId?: string;
   payload: EntitySchemaMap[T];
 };
 
-type TransactionUpdateEntity<T extends Entity> = {
+export type TransactionUpdateEntity<T extends Entity = Entity> = {
   operation: 'updateEntity';
   entityType: T;
   entityId: string;
@@ -15,7 +15,7 @@ type TransactionUpdateEntity<T extends Entity> = {
   condition?: string;
 };
 
-type TransactionAdjustEntity<T extends Entity> = {
+export type TransactionAdjustEntity<T extends Entity = Entity> = {
   operation: 'adjustEntity';
   entityType: T;
   entityId: string;
@@ -23,11 +23,17 @@ type TransactionAdjustEntity<T extends Entity> = {
   condition?: string;
 };
 
-type TransactionDeleteEntity<T extends Entity> = {
+export type TransactionDeleteEntity<T extends Entity = Entity> = {
   operation: 'deleteEntity';
   entityType: T;
   entityId: string;
 };
+
+export type TransactionOperation =
+  | TransactionCreateEntity
+  | TransactionUpdateEntity
+  | TransactionAdjustEntity
+  | TransactionDeleteEntity;
 
 export const transactional = {
   createEntity: <T extends Entity>(
@@ -41,7 +47,7 @@ export const transactional = {
       operation: 'createEntity',
       entityType,
       payload: rest as EntitySchemaMap[T],
-      ...(entityId && { entityId }),
+      ...(entityId !== undefined && { entityId }),
     };
   },
 
@@ -58,7 +64,7 @@ export const transactional = {
       entityType,
       entityId,
       payload: rest as Partial<EntitySchemaMap[T]>,
-      ...($condition && { condition: $condition }),
+      ...($condition !== undefined && { condition: $condition }),
     };
   },
 
@@ -73,7 +79,7 @@ export const transactional = {
       entityType,
       entityId,
       adjustments: rest,
-      ...($condition && { condition: $condition }),
+      ...($condition !== undefined && { condition: $condition }),
     };
   },
 
--- a/packages/react/helpers/transactional.ts
+++ b/packages/react/helpers/transactional.ts
@@ -35,6 +35,9 @@ export type TransactionOperation =
   | TransactionAdjustEntity
   | TransactionDeleteEntity;
 
+// NOTE: packages/core/helpers/transactional.ts is the server-side copy of
+// this builder. Both emit the same wire format for the execute-transaction
+// endpoint — keep operation shapes in sync when changing either file.
 export const transactional = {
   createEntity: <T extends Entity>(
     entityType: T,
--- a/packages/react/helpers/transactional.ts
+++ b/packages/react/helpers/transactional.ts
@@ -1,0 +1,88 @@
+import type { Entity, EntitySchemaMap } from '@monorise/base';
+
+type TransactionCreateEntity<T extends Entity> = {
+  operation: 'createEntity';
+  entityType: T;
+  entityId?: string;
+  payload: EntitySchemaMap[T];
+};
+
+type TransactionUpdateEntity<T extends Entity> = {
+  operation: 'updateEntity';
+  entityType: T;
+  entityId: string;
+  payload: Partial<EntitySchemaMap[T]>;
+  condition?: string;
+};
+
+type TransactionAdjustEntity<T extends Entity> = {
+  operation: 'adjustEntity';
+  entityType: T;
+  entityId: string;
+  adjustments: Record<string, number>;
+  condition?: string;
+};
+
+type TransactionDeleteEntity<T extends Entity> = {
+  operation: 'deleteEntity';
+  entityType: T;
+  entityId: string;
+};
+
+export const transactional = {
+  createEntity: <T extends Entity>(
+    entityType: T,
+    payload: EntitySchemaMap[T] & { entityId?: string },
+  ): TransactionCreateEntity<T> => {
+    const { entityId, ...rest } = payload as EntitySchemaMap[T] & {
+      entityId?: string;
+    };
+    return {
+      operation: 'createEntity',
+      entityType,
+      payload: rest as EntitySchemaMap[T],
+      ...(entityId && { entityId }),
+    };
+  },
+
+  updateEntity: <T extends Entity>(
+    entityType: T,
+    entityId: string,
+    payload: Partial<EntitySchemaMap[T]> & { $condition?: string },
+  ): TransactionUpdateEntity<T> => {
+    const { $condition, ...rest } = payload as Partial<EntitySchemaMap[T]> & {
+      $condition?: string;
+    };
+    return {
+      operation: 'updateEntity',
+      entityType,
+      entityId,
+      payload: rest as Partial<EntitySchemaMap[T]>,
+      ...($condition && { condition: $condition }),
+    };
+  },
+
+  adjustEntity: <T extends Entity>(
+    entityType: T,
+    entityId: string,
+    adjustments: Record<string, number> & { $condition?: string },
+  ): TransactionAdjustEntity<T> => {
+    const { $condition, ...rest } = adjustments;
+    return {
+      operation: 'adjustEntity',
+      entityType,
+      entityId,
+      adjustments: rest,
+      ...($condition && { condition: $condition }),
+    };
+  },
+
+  deleteEntity: <T extends Entity>(
+    entityType: T,
+    entityId: string,
+  ): TransactionDeleteEntity<T> => ({
+    operation: 'deleteEntity',
+    entityType,
+    entityId,
+  }),
+};
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -22,6 +22,8 @@ import {
   MutualDataWithIndex,
 } from './types/mutual.type';
 
+export { transactional } from './helpers/transactional';
+
 type Options<T extends Record<string, React.ComponentType<any>>> = {
   authBaseUrl?: string;
   filestoreBaseUrl?: string;
@@ -208,7 +210,6 @@ export {
 
 export default Monorise;
 
-export { transactional } from '@monorise/core';
 export { MutualDataWithIndex, MutualDataMapping, MutualData, Mutual };
 
 export type {
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -22,8 +22,6 @@ import {
   MutualDataWithIndex,
 } from './types/mutual.type';
 
-export { transactional } from './helpers/transactional';
-
 type Options<T extends Record<string, React.ComponentType<any>>> = {
   authBaseUrl?: string;
   filestoreBaseUrl?: string;
@@ -210,6 +208,7 @@ export {
 
 export default Monorise;
 
+export { transactional } from './helpers/transactional';
 export { MutualDataWithIndex, MutualDataMapping, MutualData, Mutual };
 
 export type {
@@ -219,3 +218,4 @@ export type {
   EntitySchemaMap,
   NumericFields,
 } from '@monorise/base';
+export type { TransactionOperation } from './helpers/transactional';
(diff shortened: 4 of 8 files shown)


packages/react/index.ts
```diff
@@ -22,6 +22,8 @@ import {
   MutualDataWithIndex,
 } from './types/mutual.type';
 
+export { transactional } from './helpers/transactional';
```
yeuharn96: 🔴 bug: this makes `transactional` an ambiguous star export on the `monorise` root entry — `packages/monorise/build.js` emits `export * from './core/index.js'` + `export * from './react/index.js'`, and Core exports its own `transactional` at `packages/core/index.ts:68`. Node: `SyntaxError: ... contains conflicting star exports for name 'transactional'`. TS: `TS2308`. The old form re-exported Core's symbol, so TS saw one symbol and did not error.

Move the builder to `@monorise/base` and re-export from both packages. See PR-level comment for details.

🔵 nit: also, this export sits between the import block and the `Options` type — move it down to the export block at L210 with the rest.

packages/react/helpers/transactional.ts
```diff
@@ -0,0 +1,88 @@
+import type { Entity, EntitySchemaMap } from '@monorise/base';
```
yeuharn96, line 1: 🟡 risk: L1-88 is a byte-for-byte fork of `packages/core/helpers/transactional.ts` plus a private copy of the four types from `packages/core/types/transaction.ts`. The wire format is shared with `execute-transaction.controller.ts`, so any future change to an operation shape has to land in two files or clients silently emit payloads the server rejects.

`@monorise/base` is the correct home: zod-only peer dep, already a React peer dep, already star-exported by the root index. Put `transactional` + the transaction types there and have Core and React both `export { transactional } from '@monorise/base'`.

packages/react/helpers/transactional.ts
```diff
@@ -0,0 +1,88 @@
...
+    payload: EntitySchemaMap[T] & { entityId?: string },
+  ): TransactionCreateEntity<T> => {
+    const { entityId, ...rest } = payload as EntitySchemaMap[T] & {
+      entityId?: string;
+    };
+    return {
+      operation: 'createEntity',
+      entityType,
+      payload: rest as EntitySchemaMap[T],
+      ...(entityId && { entityId }),
```
yeuharn96: 🔵 nit: `entityId && { entityId }` drops `entityId: ''`. Same for `$condition` at L61 and L76. Inherited from Core, so not introduced here — but if the logic moves to `@monorise/base` it is a one-line fix worth taking: `entityId !== undefined`.

packages/react/helpers/transactional.ts
```diff
@@ -0,0 +1,88 @@
...
+  condition?: string;
+};
+
+type TransactionDeleteEntity<T extends Entity> = {
+  operation: 'deleteEntity';
+  entityType: T;
+  entityId: string;
+};
+
+export const transactional = {
```
yeuharn96, line 41: 🔵 nit: none of the four operation types are exported, so `transaction()` in `packages/react/services/core.service.ts:279-291` still declares its own inline structural type for `operations`. Export a `TransactionOperation` union (as Core does) and type the service param with it — the two shapes can then no longer diverge.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude**

Root-cause diagnosis is right — `export { transactional } from '@monorise/core'` in `packages/react/index.ts` was the only thing dragging AWS SDK / `fs` / `async_hooks` into the browser bundle (tsup only marks `zustand` external, so Core got inlined). After this change `grep -rn "@monorise/core" packages/react` is empty, so the React package is genuinely clean.

Blocking on one thing: duplicating the implementation instead of relocating it breaks the root `monorise` entrypoint.

### 🔴 Conflicting star exports on the `monorise` root entry

`packages/monorise/build.js:96-108` generates `dist/index.js` / `dist/index.d.ts` as:

```js
export * from './core/index.js';
export * from './react/index.js';
```

Core exports `transactional` (`packages/core/index.ts:68`) and React now exports a **separate declaration** with the same name. Two distinct bindings under one name in overlapping `export *` sets is an ambiguous star export:

- Runtime: `import { transactional } from 'monorise'` → `SyntaxError: The requested module './index.js' contains conflicting star exports for name 'transactional'` (verified on Node 20).
- Types: `error TS2308: Module "./core" has already exported a member named 'transactional'. Consider explicitly re-exporting to resolve the ambiguity.` (verified with the repo's `tsc`).

This is a regression at the type level: previously React re-exported Core's *symbol*, and TS treats a re-export of the same symbol as non-ambiguous, so `idx` compiled clean. Verified both cases side by side — re-export form: no error; own-declaration form: TS2308.

**Fix:** move `transactional` and `packages/core/types/transaction.ts` into `@monorise/base`, then re-export from both Core and React. `@monorise/base` has only `zod` as a peer dep (`packages/base/package.json`), is already a React peer dep, and is already star-exported by the root index — so it stays browser-safe, both packages resolve to one symbol, the ambiguity disappears, and there is no duplicated logic to drift. The changeset then needs `@monorise/base` and `@monorise/core` bumps too.

Remaining comments inline.

wengkhing:
Addressed in 46e4045.

- Kept independent Core and React transaction builders; React no longer imports Core.
- Made the unified root entry namespace-only, so `monorise/core` and `monorise/react` remain separate and no flattened export can conflict.
- Exported React `TransactionOperation` and used it in the React API service.
- Preserved empty-string optional values with `!== undefined`.

Verified `npm run build` and runtime imports: root loads without `transactional`; both `monorise/core` and `monorise/react` expose their own builder.

packages/react/index.ts
wengkhing: Addressed in 46e4045. The unified root entry is now namespace-only, so Core and React can export separate transactional builders through monorise/core and monorise/react without a flattened root conflict.

packages/react/helpers/transactional.ts
wengkhing: Kept separate intentionally: React builds API payloads while Core owns server-side transaction handling. The unified root no longer flattens either module's exports, so the implementations remain isolated at monorise/react and monorise/core.

packages/react/helpers/transactional.ts
wengkhing: Addressed in 46e4045. React now exports TransactionOperation and its transaction() service accepts TransactionOperation[], keeping the builder and API request types aligned.

packages/react/helpers/transactional.ts
wengkhing: Addressed in 46e4045. The React builder now uses !== undefined for entityId and condition, preserving empty-string values.

wengkhing:
CI is green after `6200382` made the eventual-consistency assertion wait for mutual replication. All review threads have been addressed and resolved; requesting re-review.

packages/monorise/build.js
```diff
@@ -92,14 +92,8 @@ packages.forEach((pkg) => {
...
-// Create the main index.js file that re-exports everything
-const mainIndexContent = `// Re-export all packages from their respective modules
-export * from './base/index.js';
-export * from './core/index.js';
-export * from './react/index.js';
-export * from './sst/index.js';
-
-// Also provide named exports for each package
+// Create a root entrypoint that exposes packages without flattening their APIs.
+const mainIndexContent = `// Provide named exports for each package
```
yeuharn96: 🔴 bug: this removes the entire flattened export surface of the published `monorise` package. `import { createEntityConfig } from 'monorise'` (and every other name from base/core/react/sst) now throws `SyntaxError: does not provide an export named ...` at link time.

`.changeset/bright-mice-hug.md` has `'monorise': patch`. On a `1.6.0` package that must be `major`, with the removal described in the changeset body. Consumers on `^1.6.0` otherwise get this silently.

See PR-level comment for the alternative that keeps this a patch.

.changeset/bright-mice-hug.md
```diff
@@ -0,0 +1,6 @@
+---
+'@monorise/react': patch
+'monorise': patch
```
yeuharn96, line 3: 🔴 bug: `monorise` is `patch` but `packages/monorise/build.js` removes every flattened root export. Bump to `major` and describe the removal here — this text is the CHANGELOG entry downstream consumers will read to find out their imports broke.

packages/react/helpers/transactional.ts
```diff
@@ -0,0 +1,94 @@
+import type { Entity, EntitySchemaMap } from '@monorise/base';
```
yeuharn96, line 1: ❓ q: accepting the decision to keep this separate from Core's copy — but the two already differ after this PR (`!== undefined` here vs `&&` at `packages/core/helpers/transactional.ts:44,61,76`), and both emit the same wire format for the same `execute-transaction` endpoint. Add a comment here and in the Core copy pointing at each other, so the next person changing an operation shape knows to change both?

packages/cli/templates/files/proxy-request.ts
```diff
@@ -39,7 +40,7 @@ export const proxyRequest = async ({
       body,
       headers: {
         ...reqHeaders,
-        'x-api-key': process.env.API_KEY ?? '',
+        'x-api-key': Resource.X_API_KEY.value,
```
yeuharn96: 🟡 risk: `Resource.X_API_KEY.value` throws when the SST resource context is absent — a plain `next dev` or `next build` prerender outside `sst dev` now hard-fails here. The previous `process.env.API_KEY ?? ''` degraded to a 401 from the API instead. For a scaffolded template that's a rough first-run experience; consider `process.env.X_API_KEY ?? Resource.X_API_KEY.value` or a try/catch fallback.

🔵 nit: `API_BASE_URL` on L11 still comes from `process.env` while the key now comes from `Resource`, even though `sst.config.ts:25` links `api` too. Pick one mechanism.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude** (round 2)

Re-reviewed at `6200382`. Verified resolved:

- Star-export ambiguity is gone — `packages/monorise/build.js` no longer flattens, so no `SyntaxError: conflicting star exports` and no `TS2308`.
- `TransactionOperation` exported and wired into `transaction()` in `packages/react/services/core.service.ts:281`. Assignability checked: `TransactionCreateEntity<'user'>` → `TransactionCreateEntity<Entity>` is fine.
- `!== undefined` guards. Also verified these are safe against the server: `transaction.service.ts:282` does `op.entityId || ulid()` and `:359` does `if (op.condition)`, so an empty string is coerced away server-side either way.
- Export moved to the bottom export block.
- `tsc -p packages/react` reports no errors in any file this PR touches (the 3 errors in `actions/core.action.ts` are pre-existing and unmodified here — though note `"build": "... || true"` in `packages/react/package.json` means they'd never fail CI anyway).

One blocker plus two follow-ups.

### 🔴 The root fix is a breaking change released as a patch

Making the root namespace-only is a defensible design — but it deletes the entire flattened export surface of the published `monorise` package (currently `1.6.0`). Everything that used to resolve at `import { X } from 'monorise'` — `createEntityConfig`, `Entity`, `createEntity`, `useEntity`, `initMonorise`, `monorise` from the SST module, and the rest of all four packages — is gone. Downstream installs on a `^1.6.0` range pick this up automatically and fail at link time with `SyntaxError: The requested module does not provide an export named 'X'`. There is no deprecation window and no runtime warning.

`.changeset/bright-mice-hug.md` marks `monorise` as `patch`. It needs to be `major`, with the removal spelled out in the changeset body so it lands in the CHANGELOG. Nothing in this repo imports from the bare `monorise` specifier and the docs commit consistently teaches subpaths, so in-repo impact is nil — that argues the design is right, not that the version label is.

If a major bump isn't wanted on this PR, the alternative is the original suggestion: keep the flattened root and put `transactional` in `@monorise/base`, re-exported by Core and React. One symbol, no ambiguity, no API removal, stays a patch.

### 🟡 `packages/monorise/index.ts` is now stale and still carries the conflict

Not in the diff, so no inline comment. It still reads:

```ts
export * from '../base';
export * from '../core';
export * from '../react';
export * from '../sst';
```

It now contradicts what `build.js` actually emits. It is never compiled (`"build": "node build.js"`, no `tsc`), so it's dead source — but it is the file a maintainer will read to understand the root entry, and it still contains the exact `transactional` conflict this PR set out to remove. Update it to match the generated output, or delete it.

### On keeping the builders separate

Your call, and I'm not re-litigating it. One data point for the record: the copies diverged inside this same PR — React now uses `!== undefined`, Core still uses `&&` at `packages/core/helpers/transactional.ts:44,61,76`. Harmless today (server coerces both), but it took one commit. Worth a comment in each file pointing at the other so the next editor knows there are two.

### 🟡 Scope

The PR now carries 9 commits / 29 files: the React fix, a 16-file docs refresh, a CLI proxy-key template change, a Core 404 fix, and a test-flake fix. The changesets are correctly split per package so releases won't be wrong, but the PR title and description still describe only the React change. Please update the description to list what's actually in here — or split the CLI and Core commits out.

The Core 404 change (`update-entity.controller.ts` + test 9.6) mirrors `execute-transaction.controller.ts` correctly and has coverage — no comments. The `expect.poll` flake fix is sound; `replicateData` drains incrementally and tolerates an empty batch, so re-invoking it per poll is safe.

wengkhing:
Branch cleaned up. PR #294 now contains only the two browser-safe transaction packaging commits and five related files. The documentation/API-key/Core changes and the unrelated mutual replication test commit were removed. Fresh CI is green.

.changeset/bright-mice-hug.md
yeuharn96, line 3: 🔴 bug: still `patch`. `packages/monorise/build.js:96` removes the flattened root export surface of a `1.6.1` package — that's a `major`. Also describe the removal in the body; this text is the CHANGELOG entry that tells downstream consumers why their `import { X } from 'monorise'` stopped resolving.

packages/monorise/build.js
yeuharn96: 🔴 bug: unchanged from round 2 — `import { createEntityConfig } from 'monorise'` and every other bare-root named import now throws at link time. Either mark `monorise` as `major` in the changeset, or keep the flattened root and move `transactional` into `@monorise/base` re-exported by Core and React, which resolves the conflict without removing API and stays a patch.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude** (round 3)

Branch cleanup is exactly right — 2 commits, 5 files, nothing but the stated change. Much easier to review.

I diffed the previous head `6200382` against `304f646` across every file I'd commented on: the only differences are `packages/monorise/CHANGELOG.md` and `package.json` picking up `1.6.1` from the rebase. **Neither round-2 blocker was touched.** Restating them briefly rather than re-deriving:

1. 🔴 `.changeset/bright-mice-hug.md` still says `'monorise': patch` while `packages/monorise/build.js:96` still removes every flattened root export. `monorise` is now `1.6.1` — still 1.x, so this is a `major`. Consumers on `^1.6.0` pick it up automatically and get `SyntaxError: does not provide an export named ...` at link time for any bare `import { X } from 'monorise'`.

2. 🟡 `packages/monorise/index.ts` is still the old flattened `export *` block — now contradicting what `build.js` emits, and still containing the `transactional` conflict this PR exists to remove. Not in the diff, so no inline thread; needs updating to match the generated output, or deleting.

Re-verified at this head: `packages/core/index.ts:17` still declares its own `transactional`, so the ambiguity genuinely is resolved by the namespace-only root; nothing in the repo imports the bare `monorise` specifier; `tsc -p packages/react` is clean on all five changed files.

The cross-reference-comment suggestion on the two builder copies is optional — not blocking.

Once the changeset is `major` (or the root keeps flattening with `transactional` moved to `@monorise/base`), this is good to merge.

packages/monorise/build.js
wengkhing: Fixed in `2b569a6` — the flattened root exports are restored (`export *` from base/core/react/sst). The only name no longer at the root is `transactional`: both the Core and React copies export it, so the wildcard omits the ambiguous name. It's available via `monorise/core` (server) and `monorise/react` (browser).

.changeset/bright-mice-hug.md
wengkhing: Stays `patch`: the flattened root surface is restored in `2b569a6`, so nothing else is removed. The root `transactional` export was unusable in 1.6.1 — it statically pulled `@monorise/core` (AWS SDK, `fs`, `async_hooks`) into any browser bundle via `monorise/react` — so no consumer could be relying on it. The removal is now described in the changeset body.

packages/react/helpers/transactional.ts
wengkhing: Done in `1d4f9ad` — both copies now carry a NOTE pointing at each other and reminding to keep operation shapes in sync.

packages/cli/templates/files/proxy-request.ts
wengkhing: This file is no longer part of this PR — the CLI template/API-key changes were split into PR #296. Will address the `Resource` fallback there.

.changeset/bright-mice-hug.md
wengkhing: Resolved by `2b569a6` — flattened root exports are kept, so this remains a `patch`. Only `transactional` leaves the root (ambiguous between the Core and React copies, and broken for all browser consumers in 1.6.1). The removal is described in the changeset body.

packages/monorise/build.js
wengkhing: Resolved by `2b569a6` — flattened root exports restored, so bare-root named imports like `createEntityConfig` keep working. `transactional` is intentionally not re-exported at the root (ambiguous name; the 1.6.1 root export was what broke browser bundles). Use `monorise/core` or `monorise/react`.

wengkhing:
Round 4 pushed:

- `2b569a6` — flattened root exports restored; root no longer exports `transactional` (ambiguous between the Core and React copies; the 1.6.1 root export was what broke browser bundles, so removal stays a patch). Available via `monorise/core` and `monorise/react`.
- `1d4f9ad` — cross-reference NOTEs added to both transactional builders.
- Changeset body now describes the root `transactional` removal.

Verified against the packed tarball in a downstream Next.js consumer: root exports load without `transactional`, `createMutualConfig`/`StandardError`/etc. intact, react bundle has no `@monorise/*` imports, and the production webpack build compiles successfully. All review threads addressed and resolved; requesting re-review.

.changeset/bright-mice-hug.md
wengkhing: Full reasoning for keeping this a `patch` rather than `major`:

**1. Semver's definition of breaking is about *working* functionality.** A major bump is required when you remove or change behavior consumers could be relying on. The root `transactional` export in 1.6.1 could not be relied on by anyone: importing it statically pulled `@monorise/core` (AWS SDK v3, `fs`, `async_hooks`) into the module graph, so any browser consumer (the entire audience of `monorise/react` — webpack/Next.js bundles) failed at build time. There is no environment where the 1.6.1 root `transactional` actually worked: browsers fail to compile, and server consumers use `@monorise/core` / `monorise/core`, where `transactional` still exists unchanged.

**2. Removing it therefore breaks zero functioning usages.** If a consumer wrote `import { transactional } from 'monorise'` in 1.6.1, their build was already broken — this PR doesn't regress them, and the error they see after upgrading (missing export at link time) is the same class of failure they already had, now with a documented fix. A `major` would signal 'audit your code and migrate'; there is nothing to migrate because no working usage exists.

**3. The rest of the root surface is untouched.** Flattened exports were restored in `2b569a6` — `createEntityConfig`, `createMutualConfig`, `StandardError`, and every other bare-root import from 1.6.1 keeps resolving. The delta is exactly one name that was never usable.

**4. This PR is a bugfix by nature.** 1.6.1 introduced a regression (browser bundles pulling in server-only code). Restoring buildability is patch-scope; it would be odd to fix a shipped regression with a major that implies new breaking intent.

**5. The escape hatch is one line and documented.** Anyone who genuinely needs the server builder gets it from `monorise/core`; browser code uses `monorise/react`. The changeset body (the CHANGELOG entry) now says exactly this.

packages/monorise/build.js
```diff
@@ -92,7 +92,7 @@ packages.forEach((pkg) => {
   }
 });
 
-// Create the main index.js file that re-exports everything
+// Create the main index.js file that re-exports everything.
```
yeuharn96, line 95: 🔴 bug: the generated root `.d.ts` now emits `TS2308: Module "./core" has already exported a member named 'transactional'` — new in this PR, because published `react/index.d.ts` used to re-export Core's symbol rather than declare its own. Consumers with `skipLibCheck: false` fail to build even when they only import unrelated names; consumers with `skipLibCheck: true` get a clean type-check on `import { transactional } from 'monorise'` that then throws `SyntaxError` at load.

Append an explicit re-export to both generated strings — explicit exports beat star exports in ESM and TS:

```js
// mainIndexContent
export { transactional } from './core/index.js';
// mainIndexDtsContent
export { transactional } from './core/index';
```

Verified with the repo's `tsc`: TS2308 gone at `skipLibCheck: false`, and the name resolves to Core's builder at runtime rather than throwing.

.changeset/bright-mice-hug.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Keep the React transactional builder browser-safe.
+
+- `@monorise/react` no longer re-exports `transactional` from `@monorise/core`,
+  so browser bundles no longer pull in server-only Core code (AWS SDK, `fs`,
+  `async_hooks`). React exposes its own builder with the same wire format.
+- The unified `monorise` root no longer exports `transactional` (the name is
```
yeuharn96, line 11: 🟡 risk: true at runtime, not in the types — the root `.d.ts` still declares `transactional` (ambiguously), so with `skipLibCheck: true` the import type-checks and then throws at load. Reword once `build.js` disambiguates: root exports Core's server-side builder, use `monorise/react` in the browser.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude** (round 4)

**Withdrawing the major-bump request.** It applied to the namespace-only root, which `2b569a6` reverts. I pulled the published `monorise@1.6.1` tarball to check your reasoning and it holds:

- `dist/react/index.js:10992` — tsup inlined `var transactional = {…}` from `../core/helpers/transactional.ts`, so React's published JS already carries its own binding.
- `dist/core/index.js:9172` — Core's own binding.
- Root flattens both, so `import { transactional } from 'monorise'` **already throws `SyntaxError: conflicting star exports`** in 1.6.1.

Removing a root export that never worked at runtime is patch-level cleanup. Agreed, and `patch` is correct.

Also confirmed good: cross-reference NOTEs in both builders (`1d4f9ad`), changeset body now documents the removal, and `TransactionOperation` does not collide — Core's public export list (`dist/core/index.d.ts:4915`) exposes `TransactionService` and `transactional` only, so `transactional` is the sole ambiguous name. `tsc -p packages/react` is clean on every file this PR touches.

### 🔴 One thing 1.6.1 didn't have: the root `.d.ts` now emits TS2308

Published `dist/react/index.d.ts:1134` is `export { transactional } from '../core/index'` — a re-export of Core's *symbol*, so the flattened root d.ts had no ambiguity. React now declares its own, so the generated root d.ts has two distinct declarations. Reproduced with the repo's `tsc` on a three-module mock of `base`/`core`/`react`:

- `skipLibCheck: false` → `idx.d.ts(3,1): error TS2308: Module "./core" has already exported a member named 'transactional'.` Fires **even for a consumer that only imports `createMutualConfig`** — they never touch `transactional` and their build fails on your shipped `.d.ts`.
- `skipLibCheck: true` → no error at all, and `import { transactional } from 'monorise'` type-checks clean, then throws `SyntaxError` at load. Types advertise an export that cannot be imported.

That second case also makes the changeset's "the root no longer exports `transactional`" only half-true: it's gone at runtime, still advertised in the types.

**Fix** — the one TS2308 names in its own message. Append an explicit re-export to both generated files in `build.js`; an explicit export wins over star exports in ESM and in TS:

```js
export { transactional } from './core/index.js';   // mainIndexContent
export { transactional } from './core/index';      // mainIndexDtsContent
```

Verified: TS2308 gone at `skipLibCheck: false`, and at runtime the name resolves to Core's builder instead of throwing. The root entry already loads Core and SST, so it is server-only regardless — pointing root `transactional` at Core costs nothing and restores what 1.6.1 was trying to offer. Update the changeset bullet to match (root exports Core's builder; use `monorise/react` in the browser).

If you'd rather root genuinely not carry the name, the alternative is still moving the builder into `@monorise/base` and re-exporting from Core and React — one symbol, no ambiguity, no duplication.

Same issue applies to `packages/monorise/index.ts`, which is back in sync with `build.js` now — worth the same explicit line if it's ever compiled.

That's the last item from me.

packages/monorise/build.js
wengkhing: Fixed in `c074cd3`, without re-adding `transactional` to the root. The generated root `.d.ts` now enumerates Core's and React's exports minus `transactional` (shared names like `Mutual` resolve to Core's, matching star-export behavior); the runtime `index.js` keeps pure star-exports, where ESM already drops the ambiguous name. Verified with the repo's `tsc` from a consumer tsconfig (`skipLibCheck: false`): zero TS2308, all bare-root named imports type-check, and `import { transactional } from 'monorise'` is now a compile-time TS2305 — matching the runtime `SyntaxError`, so types and runtime agree that the name is removed from the root. It remains available via `monorise/core` and `monorise/react`.

On the suggestion to explicitly re-export Core's builder at the root: we're deliberately keeping the name removed. The 1.6.1 root export was unusable where it mattered (it pulled `@monorise/core` → AWS SDK/`fs` into any browser bundle), so removal breaks no functioning browser usage, and server consumers have `monorise/core`. That's why this stays a patch — see the changeset body.

.changeset/bright-mice-hug.md
wengkhing: No longer applicable — the root `.d.ts` no longer declares `transactional` at all (`c074cd3`), so the import fails at compile time (TS2305) under any `skipLibCheck` setting and at load time, consistently. The changeset text ('The unified `monorise` root no longer exports `transactional`... use `monorise/core` or `monorise/react`') is now accurate for both types and runtime.

packages/monorise/build.js
```diff
@@ -108,11 +108,50 @@ export * as sst from './sst/index.js';
...
+      }
+    }
+  }
+  if (names.size === 0) {
+    throw new Error(`Failed to enumerate exports from ${pkg}/index.d.ts`);
+  }
+  return names;
+};
+
+const coreNames = getDtsExportNames('core');
```
yeuharn96: 🔴 bug: `base` and `sst` are still `export *` but their names never enter this computation, so an explicit Core/React name silently overrides a star export that is ambiguous at runtime. Verified against the built output: `Entity` is declared here (Core's `Entity` class) while `import { Entity } from 'monorise'` throws `SyntaxError: conflicting star exports for name 'Entity'`. 1.6.1 excluded it from the types too, so this compiles-then-crashes where it previously failed to compile.

Dropping `Entity` from `coreNames` alone doesn't fix it — `export * from './base/index'` would then declare base's `Entity` while runtime still has none. Either enumerate all four packages with no `export *`, or stop Core exporting `Entity` under base's name.

packages/monorise/build.js
```diff
@@ -108,11 +108,50 @@ export * as sst from './sst/index.js';
...
+const getDtsExportNames = (pkg) => {
+  const dts = fs.readFileSync(path.join(distDir, pkg, 'index.d.ts'), 'utf-8');
+  const names = new Set();
+  const exportClause = /export\s+(?:type\s+)?\{([^}]*)\}/g;
+  let match;
+  while ((match = exportClause.exec(dts)) !== null) {
+    for (const part of match[1].split(',')) {
+      const name = part
+        .trim()
+        .split(/\s+as\s+/)
```
yeuharn96, line 148: 🟡 risk: `.split(/\s+as\s+/).pop()` keeps an inline `type` modifier as part of the name — `export { type Foo }` yields `"type Foo"`. `packages/base/dist/index.d.ts:488` already emits that form. Harmless for emission today (Core and React don't use it, and `export { type Foo } from …` is valid syntax anyway), but it breaks the set arithmetic: the moment Core emits `type Mutual` and React emits `Mutual`, the dedupe on L144-146 misses and TS2308 is back. It also blocks extending this to `base`/`sst` for the `Entity` fix.

`.replace(/^type\s+/, '')` after the `as` split.

packages/monorise/build.js
```diff
@@ -108,11 +108,50 @@ export * as sst from './sst/index.js';
 
 fs.writeFileSync(path.join(distDir, 'index.js'), mainIndexContent);
 
-// Create the main index.d.ts file
+// Create the main index.d.ts file.
+// Core and React declare different `transactional` builders, so star-exporting
+// both would emit TS2308. Runtime star-exports already drop the ambiguous name
+// (ESM semantics); enumerate Core and React exports here minus `transactional`
+// so the type surface matches the runtime surface. Names exported by both
```
yeuharn96: 🔵 nit: "Names exported by both (e.g. `Mutual`) resolve to Core's, matching star-export behavior" — star exports don't pick a winner, they drop the ambiguous name entirely (that's why `transactional` and `Entity` vanish from the runtime namespace). `Mutual` works because React's is type-only and elided by tsup, so there's no runtime collision. Worth correcting; as written it'll lead the next maintainer to assume first-wins.

packages/monorise/build.js
```diff
@@ -108,11 +108,50 @@ export * as sst from './sst/index.js';
...
+// (e.g. `Mutual`) resolve to Core's, matching star-export behavior.
+const getDtsExportNames = (pkg) => {
```
yeuharn96, line 139: 🔵 nit: only `export {…}` clauses are enumerated. A top-level `export declare const x` in React's tsc-emitted `index.d.ts` would be silently dropped from the root type surface — no error, just a missing export. None today (`packages/react/dist/index.d.ts` uses only `export {…}` / `export type {…}` / `export default`), but the `names.size === 0` guard won't catch a partial miss.

turbo.json
```diff
@@ -3,6 +3,16 @@
     "build": {
       "outputs": ["dist/**"]
     },
+    "monorise#build": {
```
yeuharn96, line 6: 🔵 nit: with this edge in place, the root `"build": "… turbo run build --force && npm run build:monorise"` builds `monorise` twice. The trailing `npm run build:monorise` can go.

yeuharn96 (changes requested):
🤖 **Reviewed by Claude** (round 5)

Ran the real build at this head (`npm run build`, clean tree at `004741c`) and checked the generated artifacts rather than a mock.

**TS2308 is fixed.** `tsc --skipLibCheck false` against the built `packages/monorise/dist/index.d.ts`, from a consumer importing only `StandardError`, is clean. And `transactional` is now genuinely absent from both surfaces — not in the generated `.d.ts`, not in the runtime namespace. Consistent. The `turbo.json` ordering is right too: `packages/monorise/package.json` declares no `@monorise/*` dependencies, so turbo had no edge to order on before.

### 🔴 The same trap, now on `Entity`

Against the built output:

```
dist/index.d.ts     → export { …, Entity, … } from './core/index';
import { Entity }   → SyntaxError: The requested module contains conflicting star exports for name 'Entity'
```

Type-checks clean under `skipLibCheck: true`, throws at load. `base/index.js` exports the `Entity` enum and `core/index.js` exports its `Entity` class (`Entity$1 as Entity`); root star-exports both, so ESM drops the name — but the new explicit Core clause wins over the star export in the type surface and re-declares it. It also resolves to Core's *class*, not the base enum a consumer reaching for `Entity` expects.

In 1.6.1 the root `.d.ts` star-exported both, so `Entity` was excluded from the types too — the import failed to compile. It now compiles and crashes. That's a worse failure mode, in the commit whose message is "align root type surface with runtime exports."

I diffed the full declared set against the runtime namespace of the built `dist/index.js`: `Entity` is the only real mismatch. The other eight (`MutualData*`, `CreatedEntity`, `DraftEntity`, `EntitySchemaMap`, `NumericFields`, `TransactionOperation`) are type-only and correctly have no runtime binding.

**Why the current approach can't close this:** `base` and `sst` are still `export *`, and their names never enter the ambiguity computation — only React is deduped against Core, and only `transactional` is force-dropped. I simulated removing `Entity` from the Core clause: it doesn't help, because `export * from './base/index'` then declares base's `Entity` while the runtime still has none. With any `export *` left in the root `.d.ts` there is no way to express "this name was dropped."

Two ways out:

1. Enumerate all four packages and drop every name exported by more than one — no `export *` at all in the generated `.d.ts`.
2. Fix it at the source, which is less total work: Core exporting its internal `Entity` class under the same name as base's `Entity` enum is the actual defect, and it predates this PR. Rename it (`CoreEntity`, `EntityModel`) and root `Entity` resolves to the base enum in both surfaces — strictly better than dropping it. Same shape as moving `transactional` to `@monorise/base`.

Rest is small stuff, inline.

packages/monorise/build.js
wengkhing: Fixed — went with enumerating all four packages, no `export *` left in the root `.d.ts`. Build.js now imports each built `dist/<pkg>/index.js` to detect runtime-ambiguous names (any name exported by more than one package) and omits them from the type surface entirely, so `Entity` and `transactional` are absent from the types exactly as ESM drops them at load. Names declared by multiple packages are claimed once (first package listed), preserving 1.6.1 resolutions like `Mutual` → Core and base's config types. Verified with the repo's `tsc` at `skipLibCheck: false`: `import { Entity } from 'monorise'` and `import { transactional } from 'monorise'` are now compile-time TS2305 (matching the runtime `SyntaxError`), the full bare-root surface compiles clean, zero TS2308.

packages/monorise/build.js
wengkhing: Fixed — `.replace(/^type\\s+/, '')` applied after the `as` split, so inline `type` modifiers (like base's `export { type AdjustmentCondition, ... }`) normalize to the bare name and set arithmetic works. All four packages are enumerated now, so base's inline-type clauses are covered too.

packages/monorise/build.js
wengkhing: Corrected — the comment now states that star exports drop ambiguous names entirely (which is why `transactional` and `Entity` vanish at runtime) and that `Mutual` only survives because React's is a type-only re-export that tsup elides, leaving Core's runtime value unopposed.

packages/monorise/build.js
wengkhing: Addressed — the parser now also collects `export declare const/let/var/function/class/interface/type/enum/namespace <name>` statements, so a future top-level declared export lands in the root type surface instead of silently going missing.

turbo.json
wengkhing: Removed — root `build` is now just `npm run clear-dist && turbo run build --force`; the turbo edge builds monorise once, in order.

yeuharn96 (commented):
🤖 **Reviewed by Claude** (round 6) — no blocking findings. Leaving this as a comment; approval is the maintainer's call.

Built at `5a58d0a` (clean tree, `npm run build`, 6/6 turbo tasks) and diffed the generated artifacts against the published `monorise@1.6.1` tarball.

**First, a correction to round 5.** I said the `Entity` trap was introduced by `c074cd3` and that 1.6.1 excluded `Entity` from its type surface. That was wrong. TS reports TS2308 for an ambiguous star export but still exposes the name, so 1.6.1's root types *did* declare `Entity` — `import { Entity } from 'monorise'` compiled there and threw at load, exactly as it did after `c074cd3`. The trap was pre-existing, not a regression. Worth acting on either way, and `acd8564` closes it.

**Verified on the built output:**

| check | result |
|---|---|
| Type surface vs runtime namespace | aligned — the only declared-but-absent names are the 17 genuinely type-only ones (`TransactionOperation`, `MutualData*`, `CreatedEntity`, condition types, …) |
| TS2308 count, `skipLibCheck: false` | **0** (published 1.6.1 has 2: `Entity` and `Mutual`) |
| `import { Entity }` / `import { transactional }` from root | now `TS2305: has no exported member` — matches runtime, no longer compiles-then-crashes |
| Root **runtime** surface vs 1.6.1 | **identical**, 78 exports, nothing lost or added |
| Root **type** surface vs 1.6.1 | −`Entity`, −`transactional` (neither was loadable at runtime in 1.6.1), +`TransactionOperation`; other 94 unchanged |
| Common imports (`StandardError`, `createEntityConfig`, `Mutual`, `useEntity`, `CoreFactory`) | resolve, `skipLibCheck: false` clean apart from Core d.ts errors that are byte-identical in 1.6.1 |

So the whole change is type-surface-only at the root, and the surface is now truthful. `Mutual` got fixed as a side effect. The rewritten comment block at L110-117 is accurate now, including the first-package-claims rule.

**Two non-blocking notes:**

1. `runtimeAmbiguous()` imports all four built `index.js` files, so root `.d.ts` generation now executes each package's module top level at build time — including React's `initMonorise()` (the zustand devtools warning in the build log comes from there). It works, but it silently couples the build to every package staying importable in bare Node; the day one of them needs a browser or SST context, the failure will point at `build.js` rather than the real cause. A `try/catch` with a message naming the package would pay for itself.
2. `export declare abstract class` isn't in the `exportDeclare` alternation — nothing hits it today.

Nothing further from me.

yeuharn96 (approved):

PR #295, opened 2026-07-30
docs: refresh product and deployment guidance

wengkhing:
## Summary

- Preserve the existing tree processor decision guidance while renaming the concept and correcting its trigger, traversal, and processor examples.
- Recommend shared `createMutualConfig` definitions and add valid condition, mutual-data, tag, React SDK, and SST v4 examples.
- Separate the rotatable Core `API_KEYS` allow-list from the application-owned backend proxy `X_API_KEY`; generated projects create and link the selected proxy secret without changing `MonoriseCore`.
- Return 404 when a named function update condition targets a missing entity, matching adjustment behavior.
- Remove unsupported Slack webhook claims and other outdated API references.

## Validation

- `npm -w @monorise/core test -- ConditionalUpdateEntityHttp.test.ts`
- `npm run docs:build`
- `openspec validate refresh-documentation --strict`

### DIFF
--- a/packages/cli/templates/files/sst.config.ts
+++ b/packages/cli/templates/files/sst.config.ts
@@ -12,10 +12,9 @@ export default $config({
   async run() {
     const { monorise } = await import('monorise/sst');
 
-    const { api } = new monorise.module.Core('core', {
+    const { api, xApiKey } = new monorise.module.Core('core', {
       allowOrigins: ['http://localhost:3000'],
     });
-    const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
 
     new sst.aws.Nextjs('web', {
       path: 'apps/web',
--- a/packages/cli/templates/files/sst.config.ts
+++ b/packages/cli/templates/files/sst.config.ts
@@ -12,9 +12,10 @@ export default $config({
   async run() {
     const { monorise } = await import('monorise/sst');
 
-    const { api, xApiKey } = new monorise.module.Core('core', {
+    const { api } = new monorise.module.Core('core', {
       allowOrigins: ['http://localhost:3000'],
     });
+    const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
 
     new sst.aws.Nextjs('web', {
       path: 'apps/web',
--- a/packages/cli/templates/files/sst.config.ts
+++ b/packages/cli/templates/files/sst.config.ts
@@ -15,6 +15,9 @@ export default $config({
     const { api } = new monorise.module.Core('core', {
       allowOrigins: ['http://localhost:3000'],
     });
+    // Must match one entry in Core's API_KEYS secret (defaults to
+    // ["secret1", "secret2"]). Override both before deploying:
+    // npx sst secret set X_API_KEY '...' && npx sst secret set API_KEYS '["..."]'
     const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
 
     new sst.aws.Nextjs('web', {
--- a/packages/cli/templates/files/sst.config.ts
+++ b/packages/cli/templates/files/sst.config.ts
@@ -15,15 +15,13 @@ export default $config({
     const { api } = new monorise.module.Core('core', {
       allowOrigins: ['http://localhost:3000'],
     });
+    const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
 
     new sst.aws.Nextjs('web', {
       path: 'apps/web',
+      link: [api, xApiKey],
       environment: {
         API_BASE_URL: api.url,
-        // Must match one of the values in the backend's API_KEYS secret
-        // (defaults to ["secret1", "secret2"] — override both with
-        // `npx sst secret set API_KEYS '["..."]'` before deploying).
-        API_KEY: 'secret1',
       },
     });
   },
--- a/www/docs/architecture.md
+++ b/www/docs/architecture.md
@@ -98,7 +98,7 @@ Define named `updateConditions` in the entity config, then send the permitted co
 `$condition` is optional for entity updates, but required for adjustments when the entity defines `adjustmentConditions`. A failed condition returns `409 CONFLICT`. See [Entities: Conditional writes](/concepts/entities#conditional-writes) for configuration examples.
 
 ::: warning Legacy `$where`
-Raw `$where` conditions are deprecated and disabled by default. Enable `allowLegacyWhere` only for trusted compatibility callers; new clients must use named conditions.
+Raw `$where` conditions are deprecated and disabled by default. Enable `allowLegacyWhere` only for trusted compatibility callers; new clients must use named conditions. A compatibility request places the raw clauses under `$where`, for example `{ "status": "confirmed", "$where": { "status": { "$eq": "pending" } } }`.
 :::
 
 ## Data layout cheat sheet
--- a/www/docs/architecture.md
+++ b/www/docs/architecture.md
@@ -105,8 +105,8 @@ Define named `updateConditions` in the entity config, then send the permitted co
 Response behavior for `PATCH`:
 
 - `200 OK`: update applied
-- `409 CONFLICT`: condition failed (`CONDITIONAL_CHECK_FAILED`); in conditional mode this also covers a missing entity
-- `404 NOT_FOUND`: entity missing in non-conditional mode only
+- `409 CONFLICT`: condition failed (`CONDITIONAL_CHECK_FAILED`); a static condition on a missing entity also lands here
+- `404 NOT_FOUND`: entity missing without a condition or while resolving a function condition
 - `400 BAD_REQUEST`: validation errors, unique-value conflicts, or an unknown/undefined condition name
 
 ::: warning Legacy `$where`
--- a/www/docs/architecture.md
+++ b/www/docs/architecture.md
@@ -63,7 +63,6 @@ The default Hono API exposes the following routes under `/core`. All entity rout
 | `PATCH` | `/entity/:entityType/:entityId` | Update entity (partial); supports optional named `$condition` |
 | `DELETE` | `/entity/:entityType/:entityId` | Delete entity |
 | `POST` | `/entity/:entityType/:entityId/adjust` | Atomic numeric adjustment (body: `{ field: delta }`) |
-| `POST` | `/transaction` | Execute entity operations atomically |
 
 ### Mutual endpoints
 
@@ -79,7 +78,13 @@ The default Hono API exposes the following routes under `/core`. All entity rout
 
 | Method | Route | Description |
 |--------|-------|-------------|
-| `GET` | `/tag/:entityType/:tagName` | Query tagged entities (`?group=...&start=...&end=...`) |
+| `GET` | `/tag/:entityType/:tagName` | Query tagged entities (`?group=...&query=...&start=...&end=...&limit=...&lastKey=...`) |
+
+### Transaction endpoint
+
+| Method | Route | Description |
+|--------|-------|-------------|
+| `POST` | `/transaction` | Execute entity operations atomically |
 
 Custom routes can be mounted under `/core/app/*` via `customRoutes` in your monorise config.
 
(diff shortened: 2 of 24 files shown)


www/docs/architecture.md
```diff
@@ -82,71 +83,23 @@ The default Hono API exposes the following routes under `/core`. All entity rout
...
-Response behavior for PATCH:
-
-- `200 OK`: update applied
-- `409 CONFLICT`: `$where` precondition failed (`CONDITIONAL_CHECK_FAILED`)
-  - in conditional mode, this also includes missing entities
-- `404 NOT_FOUND`: entity missing in non-conditional mode
-- `400 BAD_REQUEST`: validation errors and unique-value conflicts
-
-Compatibility notes:
+`$condition` is optional for entity updates, but required for adjustments when the entity defines `adjustmentConditions`. A failed condition returns `409 CONFLICT`. See [Entities: Conditional writes](/concepts/entities#conditional-writes) for configuration examples.
```
yeuharn96, line 103: 🟡 risk: the rewrite dropped the PATCH response-code matrix, and with it the one non-obvious behavior in it — **in conditional mode a missing entity returns `409 CONFLICT`, not `404 NOT_FOUND`**.

That is still true for named `$condition`: [entity.service.ts:250-265](https://github.com/monorist/monorise/blob/36003d1/packages/core/services/entity.service.ts#L250-L265) resolves the condition into a `ConditionExpression` and passes it to `updateEntity`, so DynamoDB fails the write with a conditional-check error whether the precondition is false *or* the item is absent. Callers that branch on 404 to detect a missing entity will silently break the moment they add a `$condition`.

Restore the matrix under the new heading:

```
- `200 OK`: update applied
- `409 CONFLICT`: condition failed (`CONDITIONAL_CHECK_FAILED`) — in conditional mode this also covers a missing entity
- `404 NOT_FOUND`: entity missing, non-conditional mode only
- `400 BAD_REQUEST`: validation errors, unique-value conflicts, unknown/undefined condition name
```

www/docs/architecture.md
```diff
@@ -60,9 +60,10 @@ The default Hono API exposes the following routes under `/core`. All entity rout
 | `GET` | `/entity/:entityType/unique/:field/:value` | Get entity by unique field |
 | `GET` | `/entity/:entityType/:entityId` | Get entity by ID |
 | `PUT` | `/entity/:entityType/:entityId` | Upsert entity (full replacement) |
-| `PATCH` | `/entity/:entityType/:entityId` | Update entity (partial); supports optional [`$where` conditions](#conditional-updates-where) |
+| `PATCH` | `/entity/:entityType/:entityId` | Update entity (partial); supports optional named `$condition` |
 | `DELETE` | `/entity/:entityType/:entityId` | Delete entity |
 | `POST` | `/entity/:entityType/:entityId/adjust` | Atomic numeric adjustment (body: `{ field: delta }`) |
+| `POST` | `/transaction` | Execute entity operations atomically |
```
yeuharn96: 🔵 nit: `/transaction` is in the **Entity endpoints** table, but the section intro states "All entity routes are validated by `entityTypeCheck` middleware" — this one isn't. It's registered at the app root ([setupRoutes.ts:77](https://github.com/monorist/monorise/blob/36003d1/packages/core/controllers/setupRoutes.ts#L77)), not under `/entity/:entityType`.

Move it to its own `### Transaction endpoint` section after Tag endpoints.

packages/cli/templates/files/sst.config.ts
```diff
@@ -15,15 +15,13 @@ export default $config({
     const { api } = new monorise.module.Core('core', {
       allowOrigins: ['http://localhost:3000'],
     });
+    const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
```
yeuharn96, line 21: 🟡 risk: this drops the only in-repo warning that `secret1` is a publicly documented default that must match `API_KEYS`. The deleted comment carried that; the new line carries nothing, so a generated project deploys with a known-public key and no signal at the point of edit. `deploying.md` has the ::: danger block, but nobody reads it while editing `sst.config.ts`.

Keep a one-liner:

```ts
// Must match one entry in the Core API_KEYS secret (defaults to
// ["secret1", "secret2"]). Override both before deploying:
// npx sst secret set X_API_KEY '...' && npx sst secret set API_KEYS '["..."]'
const xApiKey = new sst.Secret('X_API_KEY', 'secret1');
```

www/docs/deploying.md
```diff
@@ -46,10 +48,10 @@ The default API keys (`secret1`, `secret2`) are public knowledge. Anyone who kno
    npx sst secret set API_KEYS '["old-key", "new-key"]' --stage production
    ```
 
-2. **Update your proxy** to use the new key:
-   ```bash
-   npx sst secret set X_API_KEY 'new-key' --stage production
-   ```
+2. **Switch the backend proxy** to the new key:
+    ```bash
```
yeuharn96: 🔵 nit: fence indented 4 spaces; steps 1 and 3 use 3. Renders the same, but drop it to 3 for consistency.

yeuharn96 (changes requested):
🤖 Reviewed by Claude

Strong pass overall — I verified the rewritten claims against the source and the substantive ones hold up, including two that fix genuinely broken docs:

- `prejoins.entityPaths` now starts with the source entity. The old single-element example would have thrown: [prejoin-processor.ts:80](https://github.com/monorist/monorise/blob/36003d1/packages/core/processors/prejoin-processor.ts#L80) reads `entityPaths[index - 1]` unguarded, so `index 0` on an uncached path dereferences `undefined`. The new `[TEACHER, CLASS, STUDENT]` form is what the cache seeding at L59-64 actually expects, and `subscribes` is correctly required by the handler at L219-227.
- `item.data` → `item.mutualData` in the custom processor example is right; `PROJECTION_EXPRESSION.MUTUAL_DATA_ONLY` does not project `data`.
- `skipCache`, tag `end` exclusivity, descending order, `useTaggedEntities` `params.group` limits, `slackWebhook` being inert, `fromTableName` prerequisites, `coreService.transaction`, `Monorise.config` option names, `adjustEntity(..., { condition })`, and SST v4 all check out.

Five items below.

---

`www/docs/architecture.md` L82: 🔵 nit: tag row still reads `?group=...&start=...&end=...`. [tags.md](https://github.com/monorist/monorise/blob/36003d1/www/docs/concepts/tags.md) now documents `query`, `limit`, and `lastKey` too — match it here.

www/docs/architecture.md
wengkhing: Restored the PATCH response matrix, including the conditional missing-entity 409 behavior.

www/docs/architecture.md
wengkhing: Moved `/transaction` into its own endpoint section after the tag endpoints.

packages/cli/templates/files/sst.config.ts
wengkhing: Restored the warning beside the generated `X_API_KEY` secret, including the matching `API_KEYS` requirement and override commands.

www/docs/deploying.md
wengkhing: Normalized the code fence indentation to match steps 1 and 3.

wengkhing:
Addressed the review-summary nit as well: the architecture tag endpoint now lists `group`, `query`, `start`, `end`, `limit`, and `lastKey`.

www/docs/architecture.md
```diff
@@ -78,75 +78,40 @@ The default Hono API exposes the following routes under `/core`. All entity rout
...
+Response behavior for `PATCH`:
 
 - `200 OK`: update applied
-- `409 CONFLICT`: `$where` precondition failed (`CONDITIONAL_CHECK_FAILED`)
-  - in conditional mode, this also includes missing entities
-- `404 NOT_FOUND`: entity missing in non-conditional mode
-- `400 BAD_REQUEST`: validation errors and unique-value conflicts
-
-Compatibility notes:
+- `409 CONFLICT`: condition failed (`CONDITIONAL_CHECK_FAILED`); in conditional mode this also covers a missing entity
```
yeuharn96: 🔴 bug: "in conditional mode this also covers a missing entity" holds only for **static** conditions. For **function** conditions — the form [entities.md](https://github.com/monorist/monorise/blob/82e86f6/www/docs/concepts/entities.md) itself demonstrates with `archive: (_data) => ({ status: { $ne: 'archived' } })` — a missing entity returns **500**, not 409.

My suggested wording last round didn't distinguish the two; that was my miss. The chain:

1. [resolve-condition.ts:66-68](https://github.com/monorist/monorise/blob/82e86f6/packages/core/services/resolve-condition.ts#L66-L68) — a function condition calls `getEntityData()` to build the expression; a static one never does.
2. [entity.service.ts:263-266](https://github.com/monorist/monorise/blob/82e86f6/packages/core/services/entity.service.ts#L263-L266) — `getEntityData` calls `entityRepository.getEntity(...)`, which routes through `Entity.fromItem(resp.Item)`.
3. [Entity.ts:43-47](https://github.com/monorist/monorise/blob/82e86f6/packages/core/data/Entity.ts#L43-L47) — `fromItem(undefined)` throws `ENTITY_IS_UNDEFINED`. It never returns undefined, so the `entity?.data ?? {}` guard at step 2 is dead code.
4. [update-entity.controller.ts:60-98](https://github.com/monorist/monorise/blob/82e86f6/packages/core/controllers/entity/update-entity.controller.ts#L60-L98) — handles `ENTITY_NOT_FOUND`, `INVALID_CONDITION`, `UNIQUE_VALUE_EXISTS`, `CONDITIONAL_CHECK_FAILED`. Not `ENTITY_IS_UNDEFINED`, so it falls through to `throw err` at L105 → 500.

The throw also happens *before* the DynamoDB write, so it can't reach the `CONDITIONAL_CHECK_FAILED` → 409 path at all.

Actual matrix:

| Case | Status |
|---|---|
| No condition, entity missing | `404` (`ENTITY_NOT_FOUND`) |
| Static condition, entity missing | `409` (`CONDITIONAL_CHECK_FAILED`) |
| Function condition, entity missing | `500` (unhandled `ENTITY_IS_UNDEFINED`) |

Two ways to close this — your call which belongs in this PR:

**Docs-only** (keeps this PR scoped to docs): narrow the claim.

```
- `409 CONFLICT`: condition failed (`CONDITIONAL_CHECK_FAILED`); a static condition on a missing entity also lands here
- `404 NOT_FOUND`: entity missing, non-conditional mode
```

…and note that a function condition on a missing entity is currently unhandled.

**Fix the code** (preferable — the 500 is the real defect): add the missing branch to `update-entity.controller.ts`, mirroring what [adjust-entity.controller.ts:54-60](https://github.com/monorist/monorise/blob/82e86f6/packages/core/controllers/entity/adjust-entity.controller.ts#L54-L60) already does:

```ts
if (
  err instanceof StandardError &&
  err.code === StandardErrorCode.ENTITY_IS_UNDEFINED
) {
  c.status(httpStatus.NOT_FOUND);
  return c.json({ ...err.toJSON() });
}
```

That makes it 404, and the matrix line becomes accurate as-is minus the "also covers a missing entity" clause. Worth checking `PATCH` and `POST /adjust` agree afterward — `adjust` already returns 404 here, so `update` returning 409/500 for the same situation is an inconsistency independent of the docs.

❓ q: if you'd rather not touch controller code in a docs PR, the docs-only wording is fine by me — but the 500 is worth an issue either way.

yeuharn96 (changes requested):
🤖 Reviewed by Claude

All five items from the last pass are resolved and verified:

- `/transaction` moved to its own section ✓
- tag row now lists `query`/`limit`/`lastKey`, matching `tags.md` ✓
- `X_API_KEY` warning restored in the generated `sst.config.ts` ✓
- `deploying.md` fence indent normalized ✓
- PATCH response matrix restored — **but one line of it is inaccurate**, see inline.

The matrix wording I suggested last round was incomplete, and I traced it properly this time. Details on the inline comment; it's the only thing left.

www/docs/architecture.md
wengkhing: Fixed the backend inconsistency: `UpdateEntityController` now maps `ENTITY_IS_UNDEFINED` to 404, matching adjust behavior. Added a regression test for a function condition on a missing entity and updated the response matrix. The targeted suite passes all 40 tests.

yeuharn96 (approved):

PR #296, opened 2026-07-31
docs: refresh product and deployment guidance

wengkhing:
## Summary

- Preserve the existing tree processor decision guidance while correcting its trigger, traversal, and processor examples.
- Recommend shared `createMutualConfig` definitions and add valid condition, mutual-data, tag, React SDK, and SST v4 examples.
- Separate the rotatable Core `API_KEYS` allow-list from the application-owned backend proxy `X_API_KEY`; generated projects create and link the selected proxy secret without changing `MonoriseCore`.
- Return 404 when a named function update condition targets a missing entity, matching adjustment behavior.
- Remove unsupported Slack webhook claims and other outdated API references.

## Validation

- `npm -w @monorise/core test -- ConditionalUpdateEntityHttp.test.ts`
- `npm run docs:build`
- `openspec validate refresh-documentation --strict`

### DIFF
--- a/.changeset/curly-cats-rotate.md
+++ b/.changeset/curly-cats-rotate.md
@@ -1,7 +1,6 @@
 ---
 '@monorise/cli': patch
-'@monorise/sst': patch
 'monorise': patch
 ---
 
-Create and expose an `X_API_KEY` secret from `MonoriseCore`, and generate backend proxies that link the selected key instead of embedding it in the Next.js environment.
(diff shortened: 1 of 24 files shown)


wengkhing:
This replaces the previously stacked PR #295 with the same documentation, generated proxy-key, SST, and missing-entity changes rebased directly onto `main`. CI is green.

